PR #35, opened 2015-05-14
implement getnodecount, don't throw error for 0 nodes

sbebo:
Implements `MathProgBase.getnodecount` and deletes duplicate function `getnodecnt`.
Also, `get_node_count` shouldn't throw an error if node count is 0.

See JuliaOpt/JuMP.jl#442

### DIFF
--- a/src/CplexSolverInterface.jl
+++ b/src/CplexSolverInterface.jl
@@ -169,6 +169,7 @@ getconstrsolution(m::CplexMathProgModel) = get_constr_solution(m.inner)
 getreducedcosts(m::CplexMathProgModel) = get_reduced_costs(m.inner)
 getconstrduals(m::CplexMathProgModel) = get_constr_duals(m.inner)
 getrawsolver(m::CplexMathProgModel) = m.inner
+getnodecount(m::CplexMathProgModel) = get_node_count(m.inner)
 
 const var_type_map = Compat.@compat Dict(
   'C' => :Cont,
@@ -225,7 +226,6 @@ setquadobj!(m::CplexMathProgModel,rowidx,colidx,quadval) = add_qpterms!(m.inner,
 ######
 # Data
 ######
-getnodecnt(m::CplexMathProgModel) = @cpx_ccall(getnodecnt, Cint, (Ptr{Void},Ptr{Void}), m.inner.env.ptr, m.inner.lp)
 function getdettime(m::CplexMathProgModel)
     tim = Array(Cdouble,1)
     stat = @cpx_ccall(getdettime, Cint, (Ptr{Void},Ptr{Cdouble}), m.inner.env.ptr, tim)
--- a/src/cpx_solve.jl
+++ b/src/cpx_solve.jl
@@ -193,11 +193,7 @@ function get_basis(model::Model)
     return cbasis, rbasis
 end
 
-function get_node_count(model::Model)
-  ret = @cpx_ccall(getnodecnt, Cint, (Ptr{Void},Ptr{Void}), model.env.ptr, model.lp)
-  ret == 0 && error("Error getting node count")
-  return ret
-end
+get_node_count(model::Model) = @cpx_ccall(getnodecnt, Cint, (Ptr{Void},Ptr{Void}), model.env.ptr, model.lp)
 
 function get_num_cuts(model::Model,cuttype)
     cutcount = Array(Cint,1)


joehuchette:
[The docs](http://www-01.ibm.com/support/knowledgecenter/#!/SSSA5P_12.6.0/ilog.odms.cplex.help/refcallablelibrary/mipapi/getnodecnt.html?cp=SSSA5P_12.6.0%2F2-7-0-1-9) say that ``cpxgetnodecnt`` uses 0 as a sentinel value for an internal error. Maybe this function should return 0 whenever the problem isn't a MIP, and call ``cpxgetnodecnt`` (and throw on it returning 0) otherwise.

joehuchette:
In particular, we can use [this flag](https://github.com/JuliaOpt/CPLEX.jl/blob/f9f4ca4f7088154b9354ba171feae0682ca3df1f/src/cpx_model.jl#L4).

mlubin:
Zero seems like a reasonable value to return in any case if no branch and bound nodes have been explored.

joehuchette:
CPLEX doesn't seem to think so ¯\_(ツ)_/¯

mlubin:
Isn't that what the docs say?

joehuchette:
Would CPLEX consider solving a MIP at the relaxation using one B&B node?

sbebo:
Nope, I believe the node count in CPLEX can be 0.

joehuchette:
Sounds like a broken API then.

sbebo:
Nah, even Gurobi does the same. The node count *can* be zero, if the root node is not closed, e.g., if presolve closes the gap.

src/CplexSolverInterface.jl
```diff
@@ -169,6 +169,7 @@ getconstrsolution(m::CplexMathProgModel) = get_constr_solution(m.inner)
 getreducedcosts(m::CplexMathProgModel) = get_reduced_costs(m.inner)
 getconstrduals(m::CplexMathProgModel) = get_constr_duals(m.inner)
 getrawsolver(m::CplexMathProgModel) = m.inner
+getnodecount(m::CplexMathProgModel) = get_node_count(m.inner)
```
joehuchette: In line with my other comment, I might change this to something like
```jl
getnodecount(m::CplexMathProgModel) = m.inner.has_int ? get_node_count(m.inner) : 0
```

joehuchette:
That's what I mean, using the CPLEX interface how would you be able to distinguish an internal error from a situation when the node count is actually just zero (but everything else is fine)?

sbebo:
You don't, but then again, I guess you wouldn't like to identify an internal error by calling "getnodecnt".

src/CplexSolverInterface.jl
mlubin: Wouldn't it return 0 if there are no integers anyway?

joehuchette: Not if  ``get_node_count`` keeps the behavior that a zero returned by ``cpxgetnodecnt``should  throw an error.

mlubin:
I'm not really bothered by this behavior. Any case where I can imagine users calling ``getnodecount`` (say as part of numerical experiments), I would also expect them to check that the solution status reported is optimal.

joehuchette:
The docs say "If a solution exists, CPXgetnodecnt returns the node count." What happens if the model is infeasible?
```jl
m1 = Model(solver=CplexSolver())
...
solve(m1) # => :Infeasible
CPLEX.get_node_count(m.internalModel.inner) # 0?
```
Seems that throwing an error rather than a spurious zero is more honest.

joehuchette:
My point being: I'd rather err on the side of caution, rather than expect that of our users.

mlubin:
I'm a bit confused, even if the problem is infeasible, it's still valid to ask how many B&B nodes were explored in order to determine that the problem was infeasible. What does CPLEX return?

joehuchette:
Not sure, I don't have an infeasible problem handy that's hard enough that CPLEX actually has to branch to prove infeasibility.

joehuchette:
Not sure we can bank on CPLEX doing the smart thing, though, considering the docs pretty clear say that it won't.

mlubin:
But how are we supposed to model the internal state of CPLEX so that we can distinguish between the two different meanings of 0 nodes? I'd rather just not get in the way and put the blame on CPLEX when it returns something meaningless.

joehuchette:
I'd rather just give a meaningful answer when we can (0 if the problem isn't a MIP) and error out when we know we can't trust CPLEX.

sbebo:
So you'd rather throw a error also in the case where everything went smoothly and you just explored 0 nodes? That sounds silly.

joehuchette:
You'd be getting the right answer by pure luck, though. Any code that depends on this behavior would break when your model changes slightly and now CPLEX has to branch all of a sudden.

mlubin:
@joehuchette are you saying we should throw an error for any MIP where the reported node count is zero?

joehuchette:
Nope, no reason to throw away good information when it's available.

mlubin:
So when do we return 0 and when do we return an error?

joehuchette:
For other solvers, I mean. For CPLEX it seems that this information isn't reliable, so yes, I think we should throw an error whenever ``cpxgetnodecnt`` returns zero.

mlubin:
That means that any code which records node counts will have to have a special case to handle this error. E.g.,
```
status = solve(m)
# check that status is Optimal
try
    nodes = getnodecount(m)
catch
    # check if the solver is CPLEX.
    # If so, given that we know it's optimal, we set
    nodes = 0
end
```

sbebo:
I confirm, this looks silly to me, but whatever.

joehuchette:
Yes, any safe library code will have to work around the fact that CPLEXs API is ambiguous and will leak through. This isn't any different if we propagate the zero, though:
```jl
status = solve(m)
if isa(CplexSolver, m.solver)
    if status == :Infeasible
        error("CPLEX doesn't want to give you a correct answer here")
    else
        # verify that getnodecount won't give you garbage...somehow
    end
end
nodes = getnodecount(m)
```
But honestly I'd be more concerned about the casual user than the library writer in this case, who can dip down into the internals a bit and deal with these cases. I'd be more worried about the (way more common) end user who we can't expect to know that CPLEX will randomly give the wrong answer when their problem is infeasible or invalidated in some way.

joehuchette:
I think the best solution might be to implement our own heuristic in ``getnodecount`` that tries to sniff out if the zero is an error code or the right answer. Something like:
```jl
function MathProgBase.getnodecount(m:: CplexMathProgModel)
    ret = get_node_count(m.inner)
    if ret == 0
        if is_infeasible(m) || is_invalid(m.env) || is_invalid(m.lp)
            error("Cannot check number of nodes")
        end
    end
    ret
end  
```

mlubin:
@joehuchette, the use case above isn't necessarily library code, it could easily be research code. If we can't provide a good implementation of ``getnodecount`` which doesn't require everyone using it to wrap it in try/catch, I'd rather not provide a MathProgBase implementation.
We also need to verify the behavior on infeasible instances. Are there any infeasible miplib problems?

joehuchette:
Yep: http://miplib.zib.de/miplib2010-infeas.php
Even if it returns nonzeros, we're still relying on undefined behavior (or, rather, relying on something that goes against _defined_ behavior). It's also not just infeasible problems, it seems that it's any problem without a feasible solution, which is even trickier to check without digging into the solver-specific stuff.

sbebo:
Tried on a couple of miplib instances. Ok, that API sucks, and the documentation too.
On a (hard) infeasible instance
+ if I set a time limit (and clearly **no** solution is found within the time limit), Cplex *does* return the number of explored nodes so far (which is what I would expect in theory and what also Gurobi does, although the documentation says something different..). 
+ if I kill (ctrl+c), then I get 0 -- in addition, it looks like an "Error 217: No solution exists" is printed to output (...).

The state of the inner model then is kind of inconsistent, sometimes calling optimize again causes the optimization to restart (seemingly) in the correct way, otherwise you get garbage. Loading another model. In gurobi, killing the b&b and callling optimize again seems to work fine.
But I wouldn't want to do that anyway

joehuchette:
Thanks for playing around with this. I opened [a ticket with CPLEX](https://www.ibm.com/developerworks/community/forums/html/topic?id=e50e8f54-e77e-4d49-b7f8-a22e674e949b); let's see what comes of it, if anything.

mlubin:
Response from CPLEX:
```
The wording in the reference documentation is a bit misleading. CPXgetnodecnt() returns the number of nodes it required the number of nodes it took CPLEX to terminate. If the problem was proven infeasible then this may well be greater than zero.
In order to figure out why CPLEX stopped you should not look at the node count. Instead you should look at the value returned by CPXgetstat(). That tells you exactly why CPLEX stopped.
```

I vote for directly returning the value from ``CPXgetnodecnt`` and letting the user check the status of the solver. Checking that the solver terminated with the expected status is standard practice for interacting with MIP solvers, especially at the level of anyone who knows enough about what's going on in order to ask for the number of branch and bound nodes.

mlubin:
@joehuchette, what do you think?

PR #36, opened 2015-05-16
Check problem type after reading from file

sbebo:
Need to set `has_int` and `has_qc` when appropriate.

Somewhat related: not sure if `has_qc` should be true if only the objective is quadratic (QP, MIQP).
Both `CPXlpopt()` and `CPXqpopt()` accept the model, but I haven't checked if they actually do the same thing or maybe use a different method (possibly different defaults).


joehuchette:
Cool, thanks.

It'd be interesting to see if CPLEX treats the two differently or not: my hope would be that it would be smart enough to apply the best defaults regardless of which ``opt`` function is called, but who knows.

PR #47, opened 2015-08-04
Branch callback

sbebo:
I tried the branch callback to quickly implement a branching rule I had in mind, fixed a couple of obvious bugs (also in the example, that was not updated). Not sure everything's ok (and I doubt I'll test/use this feature further), but the example should work now, at least.


joehuchette:
LGTM, thanks. Not sure the ``println``s are needed in the example, but they don't really hurt.

sbebo:
Oops, forgot to remove them.

joehuchette:
Thanks!

PR #49, opened 2015-08-06
full_defines.jl updated

kaarthiksundar:
Regenerated the params file for CPLEX 12.6.2


src/full_defines.jl
```diff
@@ -468,11 +106,11 @@ const CPX_BASIC = convert(Cint,1)
 const CPX_AT_UPPER = convert(Cint,2)
 const CPX_FREE_SUPER = convert(Cint,3)
 const CPX_NO_VARIABLE = convert(Cint,2100000000)
-const CPX_CONTINUOUS = convert(Cchar,'C')
-const CPX_BINARY = convert(Cchar,'B')
-const CPX_INTEGER = convert(Cchar,'I')
-const CPX_SEMICONT = convert(Cchar,'S')
-const CPX_SEMIINT = convert(Cchar,'N')
+const CPX_CONTINUOUS = convert(Cint,'C')
```
joehuchette: Should be ``Cchar``, here and below.

src/full_defines.jl
```diff
@@ -561,9 +198,12 @@ const CPX_PARAM_RANDOMSEED = convert(Cint,1124)
 const CPX_PARAM_DETTILIM = convert(Cint,1127)
 const CPX_PARAM_FILEENCODING = convert(Cint,1129)
 const CPX_PARAM_APIENCODING = convert(Cint,1130)
-const CPX_PARAM_SOLUTIONTARGET = convert(Cint,1131)
+const CPX_PARAM_OPTIMALITYTARGET = convert(Cint,1131)
 const CPX_PARAM_CLONELOG = convert(Cint,1132)
 const CPX_PARAM_TUNINGDETTILIM = convert(Cint,1139)
+const CPX_PARAM_CPUMASK = convert(Cint,1144)
+const CPX_PARAM_SOLUTIONTYPE = convert(Cint,1147)
+const CPX_PARAM_SOLUTIONTARGET = convert(Cint,1148)
```
joehuchette: Unfortunate that this changed value from 12.6: http://www-01.ibm.com/support/knowledgecenter/#!/SSSA5P_12.6.0/ilog.odms.cplex.help/CPLEX/Parameters/topics/SolutionTarget.html

Will have to think about how (if) we want to support multiple versions of params at once (not that we do that now...)

kaarthiksundar: Fixed the Cchar.

joehuchette:
Thanks!

mlubin:
So what are we doing about multiple versions?

joehuchette:
Not sure. I added a note to the README, but I'm not sure how (if) there's a way to do this nicely. Maybe a conditional ``include`` based on the version number?

mlubin:
Yes, that would do the trick.

PR #57, opened 2015-11-09
allow length(vtype) < nvars

mlubin:
Mixed-integer conic problems were broken through cplex because the conic to quadratic wrapper introduces extra variables but doesn't intercept ``setvartype!``.

CC @emreyamangil

### DIFF
--- a/src/cpx_vars.jl
+++ b/src/cpx_vars.jl
@@ -164,7 +164,7 @@ function set_vartype!(model::Model, vtype::Vector{Char})
                       Ptr{Cint},
                       Ptr{Cchar}
                       ),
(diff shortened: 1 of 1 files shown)


joehuchette:
I'm confused as to how this is broken. Was ``num_var`` not returning the correct number of variables?

joehuchette:
Or, rather, it seems that ``length(type) < nvars`` should error out, and it should be the responsibility of the caller to reason about the number of CPLEX-level variables.

mlubin:
It seems reasonably well defined that calling ``setvartype!`` with a vector shorter than the total number of variables will just set the vartype of the first ``length(vtype)`` variables. It's either that or we have to rewrite the conic to quadratic wrapper.

joehuchette:
Yeah, you're right.

emreyamangil:
checked it, its working!!

Issue #54, opened 2015-10-15
Readme library path is misleading in CPLEX 12.5


Thuener:
Not a issue just an enhancement. In the readme the "/path/to/CPLEX/library" path is misleading, in cplex 12.5 the real path is the bin folder (/opt/ibm/ILOG/CPLEX_Studio125/cplex/bin/x86-64_sles10_4.1). That was the only way that worked for me. I try to use /opt/ibm/ILOG/CPLEX_Studio125/cplex/lib/x86-64_sles10_4.1/static_pic/lib and didn't worked.

I suggest also to put an example:
export LD_LIBRARY_PATH="/opt/ibm/ILOG/CPLEX_Studio125/cplex/bin/x86-64_sles10_4.1":$LD_LIBRARY_PATH

mlubin:
Good point, would you mind submitting a PR?

PR #61, opened 2015-12-23
Misleading path library

Thuener:
https://github.com/JuliaOpt/CPLEX.jl/issues/54


mlubin:
Thanks for this. These paths have changed from 12.5 to 12.6, so I'd prefer to put the new ones.

PR #66, opened 2016-03-12
wrapper for chgcolname

mlubin:
CC @chriscoey


mlubin:
@chriscoey, you should switch back to the master branch now

PR #80, opened 2016-08-21
Fire info callbacks in all states and maintain state information

yeesian:
ref: https://github.com/JuliaOpt/JuMP.jl/pull/814

### DIFF
--- a/src/CplexSolverInterface.jl
+++ b/src/CplexSolverInterface.jl
@@ -449,22 +449,32 @@ function mastercallback(env::Ptr{Void}, cbdata::Ptr{Void}, wherefrom::Cint, user
     cpxrawcb = CallbackData(cbdata, model.inner)
     if wherefrom == CPX_CALLBACK_MIP_CUT_FEAS || wherefrom == CPX_CALLBACK_MIP_CUT_UNBD
         state = :MIPSol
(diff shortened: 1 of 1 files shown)


src/CplexSolverInterface.jl
```diff
@@ -449,22 +449,32 @@ function mastercallback(env::Ptr{Void}, cbdata::Ptr{Void}, wherefrom::Cint, user
...
+
+    if model.infocb != nothing
+        cpxcb = CplexInfoCallbackData(cpxrawcb, state, wherefrom)
+        stat = model.infocb(cpxcb)
+        if stat == :Exit
+            terminate(model.inner)
+        end
+    end
+    if model.lazycb != nothing && state == :MIPSol
+        cpxcb = CplexLazyCallbackData(cpxrawcb, state, wherefrom, userinteraction_p)
```
joehuchette: Could this overwrite the previous value for ``cpxcb`` if there is both an info callback, a lazy callback, and the ``state == :MIPSol``?

yeesian: yup, it can. `cpxcb` here is a temporary alias within each of the `if` statements.

yeesian: Should I prefix them with a `local`?

joehuchette: Oh, nevermind, I missed line 461 where the callback function was actually called.

PR #82, opened 2016-09-16
Deprecation updates for Julia 0.5

odow:
This passes tests locally. 

Side note for interest. These deprecations caused a 2500% slow down in my algorithm run time because we are solving thousands of tiny LPs pertubing the RHS coefficients...

### DIFF
--- a/src/CPLEX.jl
+++ b/src/CPLEX.jl
@@ -3,6 +3,7 @@ __precompile__()
 module CPLEX
 
     using Compat
+    import Compat.String
 
     if is_apple()
         Libdl.dlopen("libstdc++",Libdl.RTLD_GLOBAL)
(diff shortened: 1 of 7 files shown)


mlubin:
Thanks @odow. Yes, deprecation warnings can cause significant slowdowns, more motivation to fix them ASAP.

tkelman:
What happens if you try non-ascii values for these?

mlubin:
@tkelman, quite possibly crashes, leads users to complain to IBM, and eventually they decide to support a CPLEX/Julia interface officially.

tkelman:
That's a pretty user-hostile way to try to get IBM to care. Shouldn't that be easy enough to test for, and add a check to prevent crashes if it's an issue?

odow:
Yeah good point. It's just an `isascii()` check but I don't have time to do right now.

tkelman:
`Compat.String` is also not a concrete type on 0.4, which might make all the types slower

tkelman:
Any update on testing this and/or fixing it with proper checks?

PR #85, opened 2016-09-28
Implement solver independent parameters

odow:
@mlubin Copied from https://github.com/JuliaOpt/Gurobi.jl/pull/65


mlubin:
These parameters should be set through ``setparameters!``, not mixed with the solver-specific parameters.

odow:
Oh okay will fix.

odow:
@mlubin how about now

mlubin:
Looks good. You have permission to merge.

PR #101, opened 2016-12-21
Fix on 0.6

mlubin:
This gets CPLEX passing its own tests on 0.5 and 0.6. It also passes all of JuMP's tests except for a nice segfault at the end...

### DIFF
--- a/src/cpx_common.jl
+++ b/src/cpx_common.jl
@@ -1,6 +1,7 @@
 # makes calling C functions a bit easier
 macro cpx_ccall(func, args...)
     f = "CPX$(func)"
+    args = map(esc,args)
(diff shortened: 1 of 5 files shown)


joehuchette:
Hopefully the segfault is only on v0.6...?

mlubin:
Yep

On Dec 21, 2016 21:36, "Joey Huchette" <notifications@github.com> wrote:

Hopefully the segfault is only on v0.6...?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/JuliaOpt/CPLEX.jl/pull/101#issuecomment-268703335>, or mute
the thread

PR #102, opened 2016-12-21
removed camelCase

### LINKED ISSUE #99: remove camelCase from JuMP functions
to be in line with recent changes to JuMP 
❌🐫

chriscoey:
fixes #99


mlubin:
https://github.com/JuliaOpt/CPLEX.jl/blob/master/examples/branch_jump.jl looks like a test that you can run (and also needs updating)

PR #105, opened 2017-01-02
fix multiple loadproblem! calls

blegat:
I first tried the `finalize_env` mechanism used in Gurobi.jl but it gave me segfault because it happened that the model that is thrown away gets gc'd second which cause it to be freed after the environment has already been freed by the new model that was gc'd first.

### DIFF
--- a/src/CplexSolverInterface.jl
+++ b/src/CplexSolverInterface.jl
@@ -20,7 +20,7 @@ function CplexMathProgModel(;options...)
         set_param!(env, string(name), value)
     end
 
-    m = CplexMathProgModel(_Model(env), nothing, nothing, nothing, nothing, nothing, nothing, NaN)
+    m = CplexMathProgModel(Model(env), nothing, nothing, nothing, nothing, nothing, nothing, NaN)
     return m
 end
 
@@ -76,6 +76,8 @@ function loadproblem!(m::CplexMathProgModel, filename::String)
 end
 
 function loadproblem!(m::CplexMathProgModel, A, collb, colub, obj, rowlb, rowub, sense)
+  # throw away old model but keep env
+  m.inner = Model(m.inner.env)
   add_vars!(m.inner, float(obj), float(collb), float(colub))
 
   neginf = typemin(eltype(rowlb))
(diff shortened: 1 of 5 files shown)


mlubin:
It seems suspicious that we would need to manually implement a reference counter. Is there a simpler fix?

blegat:
Since each problem has a reference to the environment we might just trust Julia's GC to detect when to delete `Env`

mlubin:
@blegat, see discussion at https://github.com/JuliaLang/julia/issues/3067

blegat:
@mlubin Thanks for the link.

So we know that when the Env finalizer is called, it is no longer referenced any where so the user will not create new models or use any models and the finalizer of all models have either already been called or are going to be called soon. Is see two solutions
* Have a counter in `Env` and a `Bool` saying whether the GC has already called finalized.
   * In the Env finalizer, if the counter is 0, free it otherwise, just set the `Bool` to `true`.
   * In the Model finalizer, free the model and ask the env to decrease the counter. If the counter goes to zero and the `Bool` is `true`, free the env.
* Have a list of models in `Env` and a `Bool` in `Model` to say whether the Model has already been freed
   * In the Env finalizer, for every problem in the list, free it then free the Env.
   * In the Model finalizer, free it then remove it from `Env`'s list.

Which one do you prefer ?

mlubin:
I'm a bit confused here. Would the issue be resolved by the approach in [Gurobi.jl](https://github.com/JuliaOpt/Gurobi.jl#reusing-the-same-gurobi-environment-for-multiple-solves) where you can optionally provide an Env to the Model object, in which case the model's finalizer won't free the Env? We can tell people to do this if they need to call ``loadproblem!`` more than once.

blegat:
Who is going to free the `Env` then ?

mlubin:
The finalizer for the ``Env``, as it should.

blegat:
But then what prevents `Env` to be finalized before the models ?
In gurobi, the `Env` does not call the finalizer, it is commented
```
# finalizer(env, free_env)  ## temporary disable: which tends to sometimes caused warnings
```

mlubin:
Good question, I didn't know that was disabled. Ok, let's go ahead with the manual reference counting. Whichever approach you prefer.

blegat:
I have implemented the first one :)

PR #107, opened 2017-01-04
Add CPX_PARAM_REDUCE

blegat:


mlubin:
What is this? ``setparameters!`` is only for the solver-independent parameters listed [here](http://mathprogbasejl.readthedocs.io/en/latest/solverinterface.html#setparameters!).

Issue #106, opened 2017-01-04
LoadError: Unable to locate CPLEX installation - windows 10


parsafard:
Hi,

I have installed cplex_studio127.win-x86-64 and julia-0.5.0-win64 on Windows 10. The Environment variables are set by CPLEX, but I still get the LoadError: Unable to locate CPLEX installation. Please let me know if you have any suggestions. 

Thanks,

blegat:
Looking at the build script, the relevant part is [here](https://github.com/JuliaOpt/CPLEX.jl/blob/0fa0c7768a1c985fbdeb2d511a725ab40a104e7f/deps/build.jl#L29-L44).

Could you try to debug that part by running `Pkg.build("CPLEX")` and see what you get ?
For instance, what do you get if you run `@show ENV` ? Do you enter `if haskey(ENV,env)` ?

mlubin:
12.7 was never tested on windows, the build script may not be correct

parsafard:
Yesterday I revised the build script my self by adding "1270", but it did not work. After you changed it, I removed CPLEX and its folder and tried to build CPLEX again, but still got the same error. However, when I open the build script, it is not updated to the new one?

@show ENV
ENV = =C:=C:\WINDOWS\system32
ALLUSERSPROFILE=C:\ProgramData
APPDATA=C:\Users\mpars\AppData\Roaming
CommonProgramFiles=C:\Program Files\Common Files
CommonProgramFiles(x86)=C:\Program Files (x86)\Common Files
CommonProgramW6432=C:\Program Files\Common Files
COMPUTERNAME=DESKTOP-BNKBM42
ComSpec=C:\WINDOWS\system32\cmd.exe
CPLEX_STUDIO_BINARIES127=C:\Program Files\IBM\ILOG\CPLEX_Studio127\cplex\bin\x64_win64;C:\Program Files\IBM\ILOG\CPLEX_Studio127\opl\bin\x64_win64;C:\Program Files\IBM\ILOG\CPLEX_Studio127\opl\oplide\;C:\Program Files\IBM\ILOG\CPLEX_Studio127\cpoptimizer\bin\x64_win64;C:\Program Files\IBM\ILOG\CPLEX_Studio127\opl\oplide\plugins\ilog.odms.ide.opllang.win32.win32.x86_64_12.7.0.0
CPLEX_STUDIO_DIR127=C:\Program Files\IBM\ILOG\CPLEX_Studio127
[...]

PR #108, opened 2017-01-04
Update build.jl

odow:
Closes https://github.com/JuliaOpt/CPLEX.jl/issues/106


odow:
@mlubin @parsafard untested but this should work

PR #109, opened 2017-02-01
Fix passing of string parameters

joehuchette:
cc @mlubin does this seem right?


mlubin:
No, we should not be touching ``.data``, and that's going away on 0.6 anyway. Don't know the right way off hand.

joehuchette:
Yeah, understand that this breaks on master.

PR #114, opened 2017-02-24
don't set warmstart if we don't know problem is MIP (#113)

### LINKED ISSUE #113: setwarmstart! throws error
Following https://github.com/JuliaOpt/JuMP.jl/pull/944, CPLEX is failing JuMP's tests via:
```
  CPLEX.CplexError(3003,"CPLEX Error  3003: Not a mixed-integer problem.\n")
   in set_warm_start!(::CPLEX.Model, ::Array{Int32,1}, ::Array{Float64,1}) at /home/mlubin/.julia/v0.5/CPLEX/src/cpx_model.jl:148
   in setwarmstart!(::CPLEX.CplexMathProgModel, ::Array{Float64,1}) at /home/mlubin/.julia/v0.5/CPLEX/src/CplexSolverInterface.jl:265
```
Warm starts should be ignored if CPLEX can't handle them, not cause errors.

mlubin:
This is a quick fix so that I can go ahead with releasing JuMP. JuMP calls ``setvartype!`` before ``setwarmstart!`` so it won't change behavior for JuMP users.

Closes #113

### DIFF
--- a/src/CplexSolverInterface.jl
+++ b/src/CplexSolverInterface.jl
@@ -282,7 +282,12 @@ getunboundedray(m::CplexMathProgModel) = get_unbounded_ray(m.inner)
 
 getbasis(m::CplexMathProgModel) = get_basis(m.inner)
 
-setwarmstart!(m::CplexMathProgModel, v) = set_warm_start!(m.inner, v)
+function setwarmstart!(m::CplexMathProgModel, v)
+    # This means that warm starts are ignored if you haven't called setvartype! first
+    if m.inner.has_int
+        set_warm_start!(m.inner, v)
+    end
+end
 
 addsos1!(m::CplexMathProgModel, idx, weight) = add_sos!(m.inner, :SOS1, idx, weight)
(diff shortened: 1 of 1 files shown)


joehuchette:
Might be good to have a warning if ``m.inner.has_int==false``, but otherwise LGTM

mlubin:
@joehuchette that warning would print out whenever someone uses CPLEX in JuMP to solve an LP.

joehuchette:
Only if there's a warm-start solution though?

mlubin:
If there's a variable that has a value loaded then ``setwarmstart!`` is now called for LPs. It used to not be called at all. So yes, I guess a warning could be okay but might be more confusing than helpful (it would also print whenever you solve a sequence of LPs).

joehuchette:
Another solution might be to store the warm start in the CPLEX.jl model until right before optimizing.

mlubin:
Agreed, doesn't need to block JuMP 0.16 though.

PR #120, opened 2017-03-19
expose the effortlevel parameter through CplexSolver()

mlubin:
Useful for Pajarito.

CC @chriscoey


mlubin:
@chriscoey let me know when you've tested this so I can merge

PR #121, opened 2017-03-21
Delete constraint(s), add and remove multiple warmstarts

aritrasep:


src/cpx_model.jl
```diff
@@ -174,6 +174,57 @@ function set_warm_start!(model::Model, indx::IVec, val::FVec, effortlevel::Integ
     end
 end
 
+function add_warm_start!(model::Model, x::Vector{Vector{Float64}}, efforts::Vector{Cint})
```
joehuchette: I'm not crazy about the naming here. Probably better to write this as ``set_warm_starts!``.

src/cpx_model.jl
```diff
@@ -174,6 +174,57 @@ function set_warm_start!(model::Model, indx::IVec, val::FVec, effortlevel::Integ
...
+	beg::Vector{Cint}, inds::Vector{Cint}, vals::Vector{Cdouble} = Cint[], Cint[], Cdouble[]
+	count::Int64 = 1
```
aritrasep: Sorry about that. Its just an old habit.

src/cpx_model.jl
aritrasep: Sure

joehuchette:
Sorry for the confusion, I meant ``set_warm_starts!`` (note the plural "s").

mlubin:
Why not follow the CPLEX terminology for these wrappers and call them mip starts instead of warm starts?

mlubin:
Make sure to account for renamings in other parts of the code, e.g., ``CplexSolverInterface.jl``.

aritrasep:
@mlubin Actually, I haven't tested it using JuMP.jl or MathProgBase.jl. I am using CPLEX.jl directly / independently to create, modify and solve a mip. So, these functions have not been wrapped for MathProgBase.jl.

mlubin:
I thought you had also renamed ``set_warm_start!`` to ``set_mip_start!`` (which you should do for consistency). In this case you also need to update the other parts of the code that call ``set_warm_start!`` to avoid breaking CPLEX for people who do use it through JuMP or MathProgBase.

aritrasep:
I didn't rename the previous `set_warm_start!` to `add_mip_start!` so that it doesn't break if used through `MathProgBase` or `JuMP`. I can change the previous `set_warm_start!` as well to `add_warm_start!`, however there are some conflicts in the way the previous `set_warm_start!` is implemented.

mlubin:
You have two options:
1) New functions are called ``set_warm_starts!`` and ``set_warm_start!`` is unchanged
2) New functions are called ``set_mip_starts!`` and ``set_warm_start!`` is renamed to ``set_mip_start!``

PR #125, opened 2017-04-14
WIP: adding cplex annotations to be used for bender decomposition

IssamT:
We are implementing a new package BlockJuMP to define block decompositions for MIPs. It will also  define oracles to the solve the subproblems and other useful directives. See an example here: https://realopt.github.io/BlockJuMP.jl/latest/basic.html.

Cplex is one of the solvers that is already supporting benders decomposition and we would like to use it with BlockJuMP. What do you think about it ?

### DIFF
--- a/src/CPLEX.jl
+++ b/src/CPLEX.jl
@@ -18,8 +18,10 @@ module CPLEX
     # Standard LP interface
     importall MathProgBase.SolverInterface
 
-    # BlockJuMP decomposition interface
-    importall BlockJuMP.BlockSolverInterface
+    # BlockJuMP Interface
+    if Pkg.installed("BlockJuMP") != nothing
+        importall BlockJuMP.BlockSolverInterface
+    end
 
     # exported functions
     # export is_valid,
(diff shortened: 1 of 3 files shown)


mlubin:
This is going to require a longer discussion, but a few quick comments.

1. Cool!
2. You should be aware of [StructJuMP](https://github.com/StructJuMP/StructJuMP.jl) which seemingly has a similar purpose to BlockJuMP. You're of course free to choose to make your own extension, but I would ask that you not use "JuMP" in the name of it given that none of the JuMP developers are involved and StructJuMP is an existing JuMP extension (with JuMP developers involved) having a similar purpose.
3. We're unlikely to merge a PR that introduces optional dependencies on external packages or implements anything except wrappers for the CPLEX C API or the official MathProgBase interface. ``newlongannotation``, for example, is fine but a dependency on BlockJuMP is not.

If you have the opportunity to attend, the [JuMP developers meetup](http://www.juliaopt.org/developersmeetup/) would be a good place to discuss this in depth.

PR #132, opened 2017-04-29
Tests

blegat:
* Add a test to check that the right `full_define` is used
* Refactor `@testset`
* Remove printing and replace it with tests using `isapprox`


joehuchette:
Looks good to me

PR #136, opened 2017-07-12
fix for 0.6

mlubin:
Left out of 5dfd745c2c167bdcceac9f23e8234428ea442397
@chriscoey, see if this fixes the issue


chriscoey:
yep that fixed it!

PR #137, opened 2017-07-13
disable JuMPfunctions on Julia 0.6

mlubin:
@chriscoey has been experiencing some world age issues involving CPLEX from Pajarito.
I don't believe something like ``eval(current_module(), Expr(:import,:JuMP))`` is allowed anymore, for the same reason that we got rid of default solvers. I'd propose to disable the undocumented "JuMPfunctions" on Julia 0.6, and if they're needed they can be moved into a separate package.

@joehuchette

### DIFF
--- a/src/CPLEX.jl
+++ b/src/CPLEX.jl
@@ -106,7 +106,9 @@ module CPLEX
     include("cpx_highlevel.jl")
 
     include("CplexSolverInterface.jl")
-    if isdir(Pkg.dir("JuMP"))
+    # These are undocumented JuMP extensions for CPLEX which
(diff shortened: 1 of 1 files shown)


mlubin:
Turns out this wasn't the source of the issues. But I'd still advocate to move these functions out. Has anyone tested them on 0.6?

joehuchette:
I'm fine moving this elsewhere for now.

tkelman:
It's messy for precompilation even on 0.5 to have code in src that's conditionally dependent on whether packages not listed in REQUIRE are installed.

Issue #117, opened 2017-02-24
Warm start solution silently discarded when ``has_int=false``


joehuchette:
ref https://github.com/JuliaOpt/CPLEX.jl/pull/114

odow:
Closing since we won't be fixing this. The new MOI wrapper caches warm-starts and adds them immediately prior to optimize.

Issue #52, opened 2015-09-01
range constraint (two-sided) support 


kaarthiksundar:
When I use a two sided range constraint in a JuMP Model, I get the following warning:

```
Julia Cplex interface doesn't properly support range (two-sided) constraints.
```

Should I be concerned when this pops up?

mlubin:
It should still be correct, but I wouldn't trust any dual values.
@joehuchette?

joehuchette:
Yes, the duals will probably be worthless, but otherwise it should work. I haven't really stressed this code path much, though, so let me know if you run into any issues.

odow:
Closing since we won't be fixing this. MOI+Bridges have resolved the issue from JuMP.

Issue #42, opened 2015-07-22
How to do "cplex.setout(ofstream)"?


kaarthiksundar:
When using CPLEX with the concert API, to write the log to an external file; I would do 
     cplex.setOut(outputfilestream)
How to do a similar thing in Julia?

joehuchette:
Try
```
CPLEX.set_logfile(m.env, "myoutputfile.txt")
```
where ``m`` is the ``CPLEX.Model`` object (``model.internalModel.inner`` in JuMP after a call to ``buildInternalModel(model)``).

mlubin:
Hmm, that won't work very well if using an extension like JuMPChance.

mlubin:
(which is the case)

mlubin:
Could we add a parameter for this to the MPB solver object? I.e., ``CplexSolver(logfile="...")``

joehuchette:
How about using the solve hook? 

Using the solver object is alright too if you want to go ahead and do it.

mlubin:
JuMPChance already uses the solve hook. I don't think multiple solve hooks work right now.
@kaarthiksundar, in the meantime, you can try using the trick [here](https://github.com/JuliaOpt/JuMP.jl/blob/master/FAQ.md) to redirect the solver output to a file from Julia.

Issue #76, opened 2016-08-14
solving twice gives different objective values


IssamT:
I was solving a model, modifying the objective function, and solving it again using JuMP and CPLEX.
The first time it is solved to optimality while the second time it is unfeasible.

I tried to find out where is the problem and I managed to reduce the model to a very simple one, and solving the latter twice on row gives the correct result on the first solve but a wrong result on the second solve ( without modifying anything in the model). 

This bug does not happen when using other solvers like CbcSolver

Also it doesn't happen if I write the  testmodel.lp and solve it twice on row using the standalone cplex binary (which means the problem isn't from the warm start of cplex)

```
using JuMP
using Cbc
using CPLEX

function test_model(C::Float64)
    mcs = Model(solver = CplexSolver())

    @variable(mcs, 0 <= z <= 100, Int)
    @variable(mcs, b[1:10], Bin)
    @constraint(mcs, roundconstr ,
      0 <= z - sum{0.5 * b[u], u in 1:10}/C <= 0.999)
    @objective(mcs, Max, z - sum{0.5 * b[u], u in 1:3}/C)

    print("FIRST SOLVE")
    solve(mcs)
    @show getobjectivevalue(mcs)
    print("SECOND SOLVE")
    solve(mcs)
    @show getobjectivevalue(mcs)
end

test_model(40.0)
```

IssamT:
julia> Pkg.status()

11 required packages:
 - Atom                          0.4.4
 - AutoHashEquals                0.0.10             master
 - Blosc                         0.1.6+             master
 - CPLEX                         0.1.3+             master
 - Cbc                           0.2.3
 - Clp                           0.2.2
 - FactCheck                     0.4.3              master
 - Graphs                        0.6.0
 - IJulia                        1.2.0
 - JuMP                          0.14.0+            master
 - LightGraphs                   0.6.1              414f14d8 (dirty)
53 additional packages:
 - BinDeps                       0.4.2
 - Blink                         0.3.4
 - Calculus                      0.1.15
 - Clustering                    0.6.0+             master
 - CodeTools                     0.3.1
 - Codecs                        0.2.0
 - Compat                        0.8.6
 - Conda                         0.2.3
 - DataStructures                0.4.5
 - Devectorize                   0.4.2              master
 - Distances                     0.3.2+             master
 - Distributions                 0.10.2
 - Docile                        0.5.23
 - FileIO                        0.1.0
 - ForwardDiff                   0.2.2
[...]

PR #139, opened 2017-07-21
WIP: MOI

odow:
<s>My guess is that this will change a lot by the final form, and once we play around with implementing more wrappers.</s>

This is now in fairly robust form. It passes virtually every MOI test (with the exception of some conic ones that aren't supported i.e. SOC, EXP etc).

- Warm starts are dealt with by storing a warm start vector in the `Model` object and calling immediately prior to `mipopt` solves (as suggested [here](https://github.com/JuliaOpt/CPLEX.jl/pull/114#issuecomment-282362405)). Closes https://github.com/JuliaOpt/CPLEX.jl/issues/117
- Adds a `logfile` keyword to redirect log. Closes https://github.com/JuliaOpt/CPLEX.jl/issues/42
- Supports interval sets (with a test for https://github.com/JuliaOpt/CPLEX.jl/issues/76). Closes https://github.com/JuliaOpt/CPLEX.jl/issues/52
- Re quad interface. Now wrap only the low-level functions and push everything through MOI which should clarify https://github.com/JuliaOpt/CPLEX.jl/issues/37
- Status codes follow MOI convention which clears up https://github.com/JuliaOpt/CPLEX.jl/issues/16

Still to do:
- [ ] callbacks


src/MOI/MathOptInterface.jl
```diff
@@ -0,0 +1,100 @@
...
+end
+ConstraintMapping() = ConstraintMapping(
+    Dict{LinConstrRef{LE}, Int}(),
+    Dict{LinConstrRef{GE}, Int}(),
+    Dict{LinConstrRef{EQ}, Int}(),
+    Dict{SVConstrRef{LE}, MOI.VariableReference}(),
+    Dict{SVConstrRef{GE}, MOI.VariableReference}(),
+    Dict{SVConstrRef{EQ}, MOI.VariableReference}(),
+    Dict{SVConstrRef{MOI.Interval{Float64}}, MOI.VariableReference}()
+)
```
odow: @joaquimg what are your thoughts on this approach (and shouldn't you be asleep)?

joaquimg: I like it, as you saw, I was brainstorming with myself yesterday. My only problem with the `Dict`is that after deleting a variable one would have to loop the entire `Dict`. But the `Array` has the problem of possibly growing a lot. And a linked list might be slow to access a single element.
You could keep going with dicts and I got for arrays and then we compare, the interface are similar anyways...

joaquimg:
I have a second question, I think adding constraints and variables, one by one might be slow, but I haven´t benchmarked yet... Planning to do that soon.

joaquimg:
I did the comparison for Xpress:
https://gist.github.com/joaquimg/5dbaf83e7555b94d79f4d01f8ed53119

It is a 10x difference!

can you do it for CPLEX and/or Gurobi?

src/MOI/MathOptInterface.jl
joaquimg: do you have a better name for: `getconstraintlist(map::ConstraintMapping, ::MOI.ScalarAffineFunction{Float64}, ::MOI.EqualsTo{Float64}) = map.equal_to`

src/MOI/MathOptInterface.jl
```diff
@@ -0,0 +1,100 @@
...
+end
+
+struct ConstraintMapping
+    # rows in constraint matrix
+    less_than::Dict{LinConstrRef{LE}, Int}
+    greater_than::Dict{LinConstrRef{GE}, Int}
+    equal_to::Dict{LinConstrRef{EQ}, Int}
+
+    # references to variable
+    variable_upper_bound::Dict{SVConstrRef{LE}, MOI.VariableReference}
```
mlubin: I was thinking that solvers could skip these mappings by matching up the constraint reference number with the variable reference number, since multiple bounds per variable are not allowed.

odow: But can't I add a lower bound and a separate upper bound? Just not two upper bounds?

mlubin: You can add a separate lower bound and upper bound. You can't add two upper bounds. The wrapper could store a variable attribute vector with the possible bound statuses per variable:
- No bounds
- Lower bound only
- Upper bound only
- Lower and upper bound, separately
- Lower and upper bound via `EqualTo`
- Lower and upper bound via `Interval`

You can check if a `ConstraintReference{SingleVariable,GreaterThan}` is valid by checking the bound flag of the corresponding variable (only if lower bound only, or lower and upper separately).

joaquimg: Do you think it is useful to have "No bounds" and "Lower and upper bound, separately"? I think they are just consequence of the others.

mlubin: I meant for this list to be mutually exclusive so you can have a single vector that identifies the status of the bounds.

src/MathOptInterface.jl
```diff
@@ -71,11 +72,16 @@ ConstraintMapping() = ConstraintMapping(
 
 mutable struct CplexSolverInstance <: MOI.AbstractSolverInstance
     inner::Model
+
     last_variable_reference::UInt64
     variable_mapping::Dict{MOI.VariableReference, Int}
+    variable_references::Vector{MOI.VariableReference}
```
joaquimg: Why do we need `variable_mapping` and `variable_references`, cant we just use `keys(variable_mapping)`?

joaquimg:
What about people that use CPLEX.jl directly? Should this be a new package since there are so many changes? I was thinking about first implementing a simple MOI and then update the solver low level API. Whats in you mind?

odow:
I just moved everything out of the way for now. The new ccall methods I'm prepending `cpx_`. I'll probably keep all the old methods as they were.

odow:
Okay this is in fairly robust form. @joehuchette did you have any ideas about callbacks?

joehuchette:
I don't think we've settled on how callbacks fit into MOI yet. Are they broken on this branch?

odow:
Okay I just reenabled the old code. For now the old MPB compatible `CplexSolver` has been renamed to `CplexMPBSolver`. It really makes sense to split these out into separate packages though. A `CPLEX.jl` that contains the low-level wrapper. A `MathOptInterfaceCPLEX.jl` with `CplexMOISolver()`, and `MathProgBaseCPLEX.jl` with `CplexMPBSolver()`

odow:
Closing in favor of #181

PR #146, opened 2017-10-01
LQOI

joaquimg:
Changes might be to aggressive.

 Is enough for [MathOptInterfaceCPLEX](https://github.com/JuliaOpt/MathOptInterfaceCPLEX.jl), with [LinQuadOptInterface](https://github.com/JuliaOpt/LinQuadOptInterface.jl)

How do we proceed?


joaquimg:
I removed stuff from MOI which is lives now in MathOptInterfaceCPLEX.

This is needed to have MathOptInterfaceCPLEX working. It is passing CPLEX.jl tests.
Should I test anything else?

Is the current organization ok?

cc @mlubin @odow @blegat

odow:
Callbacks are still unresolved. I'm not in a rush to have this merged. It almost seems like there should be a CPLEXBase.jl with this version of the C wrapper to go with MathOptInterfaceCPLEX.jl. Then CPLEX.jl can stay with the MPB version and callbacks.

joaquimg:
I am not very found of having both CPLEXBase.jl  and CPLEX.jl unless the second is completely discontinued and turned into legacy.

odow:
This PR, and https://github.com/JuliaOpt/CPLEX.jl/pull/139 are both pretty unwieldy.

I propose closing both, and starting again from the latest master.

That new PR would
- add the code from MathOptInterfaceCPLEX.jl
- add all of the new `cpx_xxx.jl` files
- retain all of the old code, except, where the arguments are the same, deprecate in favour of the new named functions. e.g., `read_model` becomes `cpx_readcopyprob`.
An alternative to the last point:
- the various `add_constrs!` `add_var!` methods leave a high surface area for bugs. Leave all of the function definitions, but replace the bodies with an appropriate error.

joaquimg:
I am ok with that

odow:
Closing in favor of #181

Issue #156, opened 2018-01-10
Missing parameters in cpx_params dictionaries


bbrunaud:
Hello, 
I was trying to test Cplex automatic benders decomposition from CPLEX 12.7 and I could not pass the param CPXPARAM_Benders_Strategy. I noticed that the cpx_params12.7.jl file is generated using:

``# grep "#define" cpxconst.h | grep "CPX_PARAM_" | awk '{ printf("\"%s\" => %s,\n",$2,$3) }'``

so parameters that start with CPXPARAM without the underscore are not captured. Here is a list of parameters that are currently not being captured.

```
[Atacama ~]$ grep "#define" /opt/cplex127/cplex/include/ilcplex/cpxconst.h | grep CPXPARAM
#define CPXPARAM_H 1
#define CPXPARAM_Advance 1001
#define CPXPARAM_Barrier_Algorithm 3007
#define CPXPARAM_Barrier_ColNonzeros 3009
#define CPXPARAM_Barrier_ConvergeTol 3002
#define CPXPARAM_Barrier_Crossover 3018
#define CPXPARAM_Barrier_Display 3010
#define CPXPARAM_Barrier_Limits_Corrections 3013
#define CPXPARAM_Barrier_Limits_Growth 3003
#define CPXPARAM_Barrier_Limits_Iteration 3012
#define CPXPARAM_Barrier_Limits_ObjRange 3004
#define CPXPARAM_Barrier_Ordering 3014
#define CPXPARAM_Barrier_QCPConvergeTol 3020
#define CPXPARAM_Barrier_StartAlg 3017
#define CPXPARAM_Benders_Strategy 1501
#define CPXPARAM_Benders_Tolerances_feasibilitycut 1509
#define CPXPARAM_Benders_Tolerances_optimalitycut 1510
#define CPXPARAM_Benders_WorkerAlgorithm 1500
#define CPXPARAM_ClockType 1006
#define CPXPARAM_Conflict_Algorithm 1073
#define CPXPARAM_Conflict_Display 1074
#define CPXPARAM_CPUmask 1144
#define CPXPARAM_DetTimeLimit 1127
#define CPXPARAM_DistMIP_Rampup_DetTimeLimit 2164
#define CPXPARAM_DistMIP_Rampup_Duration 2163
#define CPXPARAM_DistMIP_Rampup_TimeLimit 2165
#define CPXPARAM_Emphasis_Memory 1082
[...]

PR #157, opened 2018-01-10
Grep for parameters starting with 'CPXPARAM'

joehuchette:
ref #156


odow:
See also https://github.com/odow/CPLEX.jl/blob/MOI/src/cpx_defines/build_defines_file.jl#L14

When I get time I should follow up on https://github.com/JuliaOpt/CPLEX.jl/pull/139.

PR #162, opened 2018-02-02
Update build.jl to enable Cplex 12.8 support in linux

jac0320:
The most recent download of Cplex in linux (ubuntu) has the .so file with version number "1280".


joehuchette:
Thanks @jac0320! (and @jeanpauphilet)

PR #179, opened 2018-08-01
Fix 07

matbesancon:
Seemed like CPLEX.jl needs a refresh for 0.7, I fixed some things.
Things I couldn't get right:
- The finalizer for `Env` needs to be redefined
- One `ccall` is done incorrectly, yielding an error in the tests
- `importall` is used, I haven't got through all places where importall-ed package are used

### DIFF
--- a/src/CPLEX.jl
+++ b/src/CPLEX.jl
@@ -6,7 +6,7 @@ module CPLEX
         using Libdl
     end
 
-    if is_apple()
+    if Sys.isapple()
         Libdl.dlopen("libstdc++",Libdl.RTLD_GLOBAL)
     end
 
@@ -84,6 +84,7 @@ module CPLEX
     #        cplex_model
 
     using Compat
+    using SparseArrays: SparseMatrixCSC
 
     include("cpx_common.jl")
     include("cpx_env.jl")
@@ -119,6 +120,8 @@ module CPLEX
         try
             eval(current_module(), Expr(:import,:JuMP))
             include("JuMPfunctions.jl")
+		catch e
+			rethrow(e)
         end
     end
 end
--- a/src/CPLEX.jl
+++ b/src/CPLEX.jl
@@ -120,8 +120,8 @@ module CPLEX
         try
             eval(current_module(), Expr(:import,:JuMP))
             include("JuMPfunctions.jl")
-		catch e
-			rethrow(e)
+        catch e
+            rethrow(e)
         end
     end
 end
--- a/src/cpx_callbacks.jl
+++ b/src/cpx_callbacks.jl
@@ -1,6 +1,6 @@
 function get_best_bound(m::Model)
     objval_p = Vector{Cdouble}(1)
-    stat = @cpx_ccall(getbestobjval, Cint, (Ptr{Void}, Ptr{Void}, Ptr{Cdouble}), m.env.ptr, m.lp, objval_p)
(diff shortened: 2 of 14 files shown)


mlubin:
Thanks! We don't have CI set up so this is hard to test (I don't have a CPLEX license). Could you confirm that CPLEX passes tests on both 0.6 and 0.7 under this PR? We are not dropping 0.6 support yet.

src/CPLEX.jl
```diff
@@ -119,6 +120,8 @@ module CPLEX
         try
             eval(current_module(), Expr(:import,:JuMP))
             include("JuMPfunctions.jl")
+		catch e
```
mlubin: Looks like a whitespace mismatch

test/runtests.jl
```diff
@@ -1,4 +1,4 @@
-using Base.Test
+using Test
```
mlubin: Should this be Compat.Test?

src/cpx_callbacks.jl
```diff
@@ -19,8 +19,8 @@ function setcallbackcut(cbdata::CallbackData, where::Cint, ind::Vector{Cint}, va
     @assert length(val) == len
     sns = convert(Cint, sense)
     stat = @cpx_ccall(cutcallbackadd, Cint, (
-                      Ptr{Void},
-                      Ptr{Void},
+                      Ptr{Nothing},
```
mlubin: Please use `Cvoid` instead of `Nothing` for `void*` types.

test/runtests.jl
matbesancon: this looks fine with just "Test", it was suggested directly by the compiler

matbesancon:
As mentioned, there is still a ccall yielding an error and I could not track down where it's coming from
```
ERROR: LoadError: LoadError: syntax: ccall argument types must be a tuple; try "(T,)" and check if you specified a correct return type
Stacktrace:
 [1] include at ./boot.jl:317 [inlined]
 [2] include_relative(::Module, ::String) at ./loading.jl:1034
 [3] include at ./sysimg.jl:29 [inlined]
 [4] include(::String) at /home/mathieu/.julia/dev/CPLEX/src/CPLEX.jl:3
 [5] top-level scope at none:0
 [6] include at ./boot.jl:317 [inlined]
 [7] include_relative(::Module, ::String) at ./loading.jl:1034
 [8] include(::Module, ::String) at ./sysimg.jl:29
 [9] top-level scope at none:0
 [10] eval at ./boot.jl:319 [inlined]
 [11] eval(::Expr) at ./client.jl:394
 [12] top-level scope at ./none:3 [inlined]
 [13] top-level scope at ./<missing>:0
in expression starting at /home/mathieu/.julia/dev/CPLEX/src/cpx_solve.jl:1
in expression starting at /home/mathieu/.julia/dev/CPLEX/src/CPLEX.jl:112
```

src/CPLEX.jl
matbesancon: I got tricked by a tab-space conversion

src/cpx_callbacks.jl
matbesancon: fixed in last commit

src/CPLEX.jl
matbesancon: fixed in last commit

PR #181, opened 2018-08-05
MOI/LQOI basic support for ILP

IssamT:


src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
+export CplexOptimizer
```
odow: Refactor `CplexOptimizer` to `CPLEX.Optimizer` and remove the export.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
...
+
+using LinQuadOptInterface
+const LQOI = LinQuadOptInterface
+const MOI  = LQOI.MOI
+
+const SUPPORTED_OBJECTIVES = [
+    LQOI.Linear
```
odow: Add `LQOI.SinVar`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
...
+    env = Env()
+    return Model(env::Env)
+end
+
+function CplexOptimizer(;kwargs...)        
+    env = Env()
+    for (name,value) in kwargs
+        set_param!(env, string(name), value)
+    end
+    m = CplexOptimizer(nothing)
```
odow: Style: always name the optimizer `model`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
...
+    m.env = env
+    return m
+end
+
+LQOI.supported_constraints(s::CplexOptimizer) = SUPPORTED_CONSTRAINTS
```
odow: Style: don't name `s` it it isn't used

src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
...
+LQOI.backend_type(m::CplexOptimizer, ::MOI.EqualTo{Float64})     = Cchar('E')
+LQOI.backend_type(m::CplexOptimizer, ::MOI.LessThan{Float64})    = Cchar('L')
+LQOI.backend_type(m::CplexOptimizer, ::MOI.GreaterThan{Float64}) = Cchar('G')
+
+LQOI.backend_type(m::CplexOptimizer, ::MOI.Zeros)                = Cchar('B')
+LQOI.backend_type(m::CplexOptimizer, ::MOI.Nonpositives)         = Cchar('L')
+LQOI.backend_type(m::CplexOptimizer, ::MOI.Nonnegatives)         = Cchar('E')
+
+# TODO - improve single type
+function LQOI.change_variable_bounds!(instance::CplexOptimizer,
```
odow: Style: refactor `instance` to `model`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
...
+LQOI.backend_type(m::CplexOptimizer, ::MOI.LessThan{Float64})    = Cchar('L')
+LQOI.backend_type(m::CplexOptimizer, ::MOI.GreaterThan{Float64}) = Cchar('G')
+
+LQOI.backend_type(m::CplexOptimizer, ::MOI.Zeros)                = Cchar('B')
+LQOI.backend_type(m::CplexOptimizer, ::MOI.Nonpositives)         = Cchar('L')
+LQOI.backend_type(m::CplexOptimizer, ::MOI.Nonnegatives)         = Cchar('E')
+
+# TODO - improve single type
+function LQOI.change_variable_bounds!(instance::CplexOptimizer, 
+        colvec::Vector{Int}, valvec::Vector{Float64}, sensevec::Vector{Cchar})
```
odow: Style: use English understandable names. `columns` instead of `colvec` and so on.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
...
+function LQOI.change_linear_constraint_sense!(instance::CplexOptimizer, 
+        rows::Vector{Int}, senses::Vector{Cchar})
+                                              
+    chg_sense!(ivec(rows), senses)
+end
+
+function LQOI.set_linear_objective!(instance::CplexOptimizer, 
+        columns::Vector{Int}, coefficients::Vector{Float64})
+    
+    n = num_var(instance.inner)
```
odow: Style: no one-letter variable names

src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
...
+    stat = get_solution_info(instance.inner)[4]
+    if stat == 1
+        return MOI.FeasiblePoint
+    else
+        return MOI.UnknownResultStatus
+    end
+end
+
+function LQOI.get_variable_primal_solution!(instance::CplexOptimizer, result)
+    fill_solution(instance.inner, result)
```
odow: Nit: `get_` instead of `fill_`?

src/MOIWrapper.jl
```diff
@@ -0,0 +1,283 @@
...
+function LQOI.get_variable_dual_solution!(instance::CplexOptimizer, place)
+    fill_reduced_costs(instance.inner, place)
+end
+
+function LQOI.get_linear_dual_solution!(instance::CplexOptimizer, place)
+    fill_constr_duals(instance.inner, place)
+end
+
+function LQOI.get_objective_value(instance::CplexOptimizer)
+    get_objval(instance.inner)
```
odow: Style: all functions that explicitly return something should use `return`

src/cpx_constrs.jl
```diff
@@ -156,6 +166,21 @@ function get_constr_senses(model::Model)
     return senses
 end
 
+function chg_sense!(model::Model, indices::IVec, senses::CVec)
```
odow: Style: I think if we are wrapping new CPLEX functions, we should use exactly the CPLEX name. So refactor `chg_sense!` to `chgsense`.

src/cpx_vars.jl
```diff
@@ -78,6 +78,22 @@ function add_var!(model::Model, constridx::Vector, constrcoef::Vector, l::Vector
     return add_var!(model, ivec(constridx), fvec(constrcoef), fvec(l), fvec(u), fvec(objcoef))
 end
 
+function get_varLB(model::Model, col::Cint)
```
odow: Style: `getlb`

src/MOIWrapper.jl
IssamT: Since there are already get functions with the same name, the name should be different for inplace functions. following our discussion yesterday, I will use `get_` with `!`

src/MOIWrapper.jl
IssamT: But `n` and `m` are often used in linear programming. And these are local vars in functions that always fully fit in a laptop screen; so if one doesn't know what the variable is, he can just look above.

src/cpx_constrs.jl
IssamT: I only added few functions, so I tried to keep the style consistent with the rest of the wrapper functions. Plus I think if we use underscores everywhere else in julia code, it's better to keep those underscores. I'll open an issue to see what the others think of this.

src/cpx_vars.jl
IssamT: I think we should keep the name `get_lb` for the function that does exactly the same as `cpxgetlb`, namely getting lowerbounds for a range of columns and not for just one column.

src/MOIWrapper.jl
IssamT: done

IssamT:
@odow, are you fine with the current names?

src/cpx_constrs.jl
```diff
@@ -165,12 +192,31 @@ function set_constr_senses!(model::Model, senses::Vector)
...
                       ),
-                      model.env.ptr, model.lp, ncons, Cint[0:ncons-1;], convert(Vector{Cchar},senses))
+                      model.env.ptr, model.lp, ncons, Cint[0:ncons-1;], 
+                      convert(Vector{Cchar},senses))
     if stat != 0
         throw(CplexError(model.env, stat))
     end
 end
 
+function c_api_getrhs(model::Model, row_start::Cint, row_end::Cint)
```
odow: I think users should have to pass in the `rhs` vector as well.
```julia
function c_api_getrhs(model::Model, rhs::Vector{Cdouble}, row_start::Cint, row_end::Cint)
```

IssamT: @odow, I agree with you. higher layers can benefit from that doing in place modification even if it's not the case of LQOI for this specific function. I will update it.

IssamT:
How about renaming all files like

cpx_model.jl -> c_api_model.jl
cpx_vars.jl -> c_api_vars.jl

and move the helper functions that we want to keep as few as possible from those file to a single file

helpers_c_api.jl

IssamT:
@odow should we go with the above renaming?

odow:
I think for now, leave the current cplex unmodified and place any new functions in a new file. We can make another pr that does a tidy

IssamT:
I also renamed some of the old ones used also by Mathprogbase when they matched the C API. If you prefer doing the renaming in another PR, then I think it's better to keep the functions in the same files. otherwise we will have subsets of functions organized differently without any reason.

src/cpx_constrs.jl
```diff
@@ -130,14 +140,15 @@ function add_rangeconstrs!(model::Model, A::CoeffMat, lb::Vector, ub::Vector)
...
-function num_constr(model::Model)
+function c_api_getnumrows(model::Model)
     ncons = @cpx_ccall(getnumrows, Cint, (
                        Ptr{Void},
                        Ptr{Void}
                        ),
                        model.env.ptr, model.lp)
     return ncons
 end
+num_constr = c_api_getnumrows # Needed by MathProgBase Interface
```
odow: I initially missed this, it might be better to go `num_constr(model::Model) = c_api_getnumrows(model)`

src/cpx_constrs.jl
```diff
@@ -165,12 +192,29 @@ function set_constr_senses!(model::Model, senses::Vector)
...
-                      model.env.ptr, model.lp, ncons, Cint[0:ncons-1;], convert(Vector{Cchar},senses))
+                      model.env.ptr, model.lp, ncons, Cint[0:ncons-1;], 
+                      convert(Vector{Cchar},senses))
     if stat != 0
         throw(CplexError(model.env, stat))
     end
 end
 
+function c_api_getrhs(model::Model, rhs::Vector{Cdouble}, row_start::Cint, row_end::Cint)
+    ncons = 1
```
odow: Not used?

src/cpx_solve.jl
```diff
@@ -79,28 +80,53 @@ function get_objval(model::Model)
   end
   return objval[1]
 end
+get_objval = c_api_getobjval
```
odow: `get_objval(model::Model) = c_api_getobjval(model)`

src/cpx_solve.jl
```diff
@@ -79,28 +80,53 @@ function get_objval(model::Model)
...
 
-function get_solution(model::Model)
-  nvars = num_var(model)
-  x = Vector{Cdouble}(nvars)
+function c_api_solninfo(model::Model)    
+  solnmethod_p = [Cint(-1)]
```
odow: You can make these `solnmethod_p = Ref{Cint}()`, and then return `solnmethod_p[]`. This avoids allocating a vector.

src/cpx_solve.jl
```diff
@@ -285,4 +326,9 @@ const status_symbols = Dict(
 )
 
 get_status(model::Model) = status_symbols[Int(get_status_code(model))]::Symbol
-get_status_code(model::Model) = @cpx_ccall(getstat, Cint, (Ptr{Void}, Ptr{Void}), model.env.ptr, model.lp)
+
+function c_api_getstat(model::Model) 
+    return @cpx_ccall(getstat, Cint, (Ptr{Void}, Ptr{Void}), 
+                      model.env.ptr, model.lp)
+end
+get_status_code = c_api_getstat
```
odow: `get_status_code(model::Model) = c_api_getstat(model)`

src/cpx_vars.jl
```diff
@@ -78,6 +78,23 @@ function add_var!(model::Model, constridx::Vector, constrcoef::Vector, l::Vector
     return add_var!(model, ivec(constridx), fvec(constrcoef), fvec(l), fvec(u), fvec(objcoef))
 end
 
+function c_api_getlb(model::Model, col_start::Cint, col_end::Cint)
+    lb = Vector{Cdouble}(col_end - col_start + 1)
```
odow: Make this in-place like `getrhs`.

src/cpx_vars.jl
```diff
@@ -191,14 +255,15 @@ function get_vartype(model::Model)
...
-function num_var(model::Model)
+function c_api_getnumcols(model::Model)
     nvar = @cpx_ccall(getnumcols, Cint, (
                       Ptr{Void},
                       Ptr{Void}
                       ),
                       model.env.ptr, model.lp)
     return(nvar)
 end
+num_var = c_api_getnumcols
```
odow: Here again.

test/MOIWrapper.jl
```diff
@@ -0,0 +1,36 @@
+using CPLEX, Base.Test, MathOptInterface, MathOptInterface.Test
```
odow: You should test this on v0.7 as well. It will need `Compat.Test` instead of `Base.Test`, or we can leave that for another PR.

test/MOIWrapper.jl
```diff
@@ -0,0 +1,36 @@
...
+                "linear12", "linear8a","linear8b","linear8c"])
+    end
+end
+
+@testset "Integer Linear tests" begin
+    intconfig = MOIT.TestConfig()
+    solver = CPLEX.Optimizer()
+    MOIT.intlineartest(solver, intconfig, ["int2", "int3"]) 
+    # 3 is ranged, 2 has sos
+end
```
odow: Go see the Gurobi and GLPK tests, there are a few missing here.

src/cpx_model.jl
```diff
@@ -52,20 +52,22 @@ end
...
+function get_sense(model::Model) 
+    sense_int = c_api_getobjsen(model)
     if sense_int == 1
-        sense = :Min
-    elseif sense_int == -1
-        sense = :Max
-    else
-        error("CPLEX: problem object or environment does not exist")
+        return :Min
+    else
```
odow: Why did we drop the check for `sense_int == -1` and the `else` error?

src/MOIWrapper.jl
```diff
@@ -0,0 +1,280 @@
...
+    return c_api_getpi(model.inner, place)
+end
+
+function LQOI.get_objective_value(model::Optimizer)
+    return c_api_getobjval(model.inner)
+end
+
+function MOI.free!(m::Optimizer)
+    # close_CPLEX(model.inner)   
+end
```
odow: This method is not needed.

IssamT:
I think we can keep the upgrade to julia 0.7 for another PR, since the MathProgBase interface is also failing it.

test/MOIWrapper.jl
```diff
@@ -0,0 +1,60 @@
...
+    solver = CPLEX.Optimizer()
+    MOIT.intlineartest(solver, intconfig, ["int2", "int3"]) 
+    # 3 is ranged, 2 has sos
+    @testset "int3" begin
+        MOIT.int3test(
+            MOIB.SplitInterval{Float64}(CPLEX.Optimizer()),
+            MOIT.TestConfig()
+        )
+    end
+end
```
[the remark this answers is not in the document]
IssamT: Done

test/MOIWrapper.jl
```diff
@@ -57,4 +57,24 @@ end
             MOIT.TestConfig()
         )
     end
+end
+
+@testset "ModelLike tests" begin
+    solver = CPLEX.Optimizer()
+    # TODO add a test config to disable testing problem name.
+    # @testset "nametest" begin
+    #     MOIT.nametest(solver)
```
odow: Why does the nametest fail? Since CLPEX uses LQOI it should pass...

IssamT: You're right. My bad.

IssamT:
I think it s better to do the update for MOI 0.6 in a separate PR, to keep a version for comparaison. Any minor issues left can be discussed in the MOI 0.6 update PR.

odow:
Okay, I have tagged a release of CPLEX prior to merging this: https://github.com/JuliaLang/METADATA.jl/pull/17661

I'll squash and merge this, then you can make another PR updating to MOI v0.6 and Julia v0.7. Then we can tag as v0.4.0

PR #183, opened 2018-09-11
Minor updates for moi06

IssamT:
Tests are passing with LQOI master. As JuliaOpt/GLPK.jl#74, it can be merged once LQOI is tagged.
Next step will be to update to julia 1.0


joaquimg:
Can we merge?

PR #184, opened 2018-09-12
Fixes for julia 1.0

IssamT:
This is a continuation on the nice work started by @matbesancon in #179. Tests are now passing for 0.6 and 1.0. Still some work to do:

- [x] pass conic tests. I temporarily disabled them since MPB still uses the deprecated `speye`.
- [ ] fix `cpx_ccall_intercept`. I temporarily replaced it with `cpx_ccall`.
- [x] fix finalizers.

PR #188, opened 2018-10-10
Update build.jl to Julia 1.0

bkamins:
Some methods changed names and we hit a case of https://discourse.julialang.org/t/another-possible-solution-to-the-global-scope-debacle/15894 debate.

If it goes through CI it would be great if the release were tagged as I am teaching this soon. Thanks 😄.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -15,7 +15,7 @@ function write_depsfile(path)
     end
 end
 
-if is_apple()
+if Sys.isapple()
     Libdl.dlopen("libstdc++",Libdl.RTLD_GLOBAL)
 end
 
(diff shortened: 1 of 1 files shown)


odow:
There is a work in progress for Julia 1.0 at https://github.com/JuliaOpt/CPLEX.jl/pull/184.

> If it goes through CI it would be great if the release were tagged as I am teaching this soon. Thanks 😄.

There is no CI because of the license issues. I'm also not sure what the time-line is on releasing CPLEX.jl for v1. You might want to explore teaching on Julia v0.6 or a different solver just in case.

bkamins:
OK - I now see it is covered in #184 so closing.

bkamins:
@odow Thank you for a quick response!

PR #192, opened 2018-10-16
fixing broadcast in function that sets branching priorities

vitornesello:


src/cpx_solve.jl
```diff
@@ -32,7 +32,7 @@ function set_branching_priority(model::Model, indices::Vector{Cint}, priority::V
                       Ptr{Cint},
                       Ptr{Cint},
                       Ptr{Cint}),
-                      model.env.ptr, model.lp, cnt, indices-Cint(1), priority, direction)
+                      model.env.ptr, model.lp, cnt, indices.-Cint(1), priority, direction)
```
blegat: Add spaces before and after the binary operator to satisfy [JuMP style guidelines](http://www.juliaopt.org/JuMP.jl/latest/style.html#Whitespace-1)

PR #198, opened 2018-11-10
Malloc error

chengg04:
Two files I have added: "./src/cpxncb.jl" and "./examples/admipex9_cb.jl". When running the "admipex9_cb.jl" file, I get the following error message:

```
include("/Users/chengguo/.julia/v0.6/CPLEX/examples/admipex9_cb.jl")
julia(8417,0x7fffb95f83c0) malloc: *** error for object 0x7fb8a6143c90: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug

signal (6): Abort trap: 6
while loading /Users/chengguo/.julia/v0.6/CPLEX/examples/admipex9_cb.jl, in expression starting on line 44
__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)
Allocations: 2529923 (Pool: 2528636; Big: 1287); GC: 3
Abort trap: 6
```

### DIFF
--- a/examples/admipex9_cb.jl
+++ b/examples/admipex9_cb.jl
@@ -1,18 +1,19 @@
 #callbacks for cap problem
-using CPLEX
+using CPLEX, Revise#, Gallium
 import Base: convert, unsafe_convert, show, copy
 using Compat
+import Gallium
+#revise julia
+# JuliaPath="/Users/chengguo/.julia/v0.6/CPLEX/src/"
+# include("/Users/chengguo/.julia/v0.6/CPLEX/deps/deps.jl")
 
-JuliaPath="/Users/chengguo/.julia/v0.6/CPLEX/src/"
-include("/Users/chengguo/.julia/v0.6/CPLEX/deps/deps.jl")
-
-include(string(JuliaPath,"cpx_common.jl"))
-include(string(JuliaPath,"cpx_env.jl"))
-include(string(JuliaPath,"cpx_model.jl"))
-include(string(JuliaPath,"full_defines_1280.jl"))
-include(string(JuliaPath,"cpx_params_1280.jl"))
-include(string(JuliaPath,"cpx_callbacks.jl"))
-include(string(JuliaPath,"cpx_newcbs.jl"))
+# include(string(JuliaPath,"cpx_common.jl"))
+# include(string(JuliaPath,"cpx_env.jl"))
+# include(string(JuliaPath,"cpx_model.jl"))
+# include(string(JuliaPath,"full_defines_1280.jl"))
+# include(string(JuliaPath,"cpx_params_1280.jl"))
+# include(string(JuliaPath,"cpx_callbacks.jl"))
+# include(string(JuliaPath,"cpx_newcbs.jl"))
 # Libdl.dlopen("libstdc++",Libdl.RTLD_GLOBAL)
 ##################
 #example of cplex callable library, admipex9.c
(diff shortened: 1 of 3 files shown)


chengg04:
Erm...maybe this information will be useful: the `rounddownheur` is based on this CPX callable library function: 

```
CPXLIBAPI
int CPXPUBLIC
   CPXcallbackpostheursoln (CPXCALLBACKCONTEXTptr context, int cnt,
                            int const *ind, double const *val,
                            double obj,
                            CPXCALLBACKSOLUTIONSTRATEGY strat);
```
The `int const *` bugs me. I suspect the `Base.unsafe_converter` function does not transfer a reference properly to a constant pointer.

chriscoey:
Asking on the Julia Slack can be helpful for things like this, if you find the right channel

chengg04:
> Asking on the Julia Slack can be helpful for things like this, if you find the right channel

Thanks! I just joined the Slack. And "#helpdesk" looks like the right channel :)

src/cpx_newcbs.jl
```diff
@@ -0,0 +1,211 @@
...
+        throw(CplexError(model.env, ret))
+    end
+    # We need to keep a reference to the callback function so that it isn't
+    # garbage collected.
+    model.callback = user_handle
+    return
+end
+
+# int CPXXcallbackabort( CPXCALLBACKCONTEXptr context )
+function cpx_callbackabort(callback_data::CallbackContext)
```
odow: Okay @chengg04, I've wrapped `callbackabort` here so you can see what I did. 

Now we just need to do the same for the other functions here:
https://www.ibm.com/support/knowledgecenter/en/SSSA5P_12.8.0/ilog.odms.cplex.help/refcallablelibrary/groups/homepagecallbacks.html

examples/admipex9_cb.jl
```diff
@@ -0,0 +1,21 @@
...
+
+function my_callback(cb_context::CPLEX.CallbackContext, context_id::Clong)
+    println("Aborting!")
+    CPLEX.cpx_callbackabort(cb_context)
+end
+
+context_id = Clong(0) | CPLEX.CPX_CALLBACKCONTEXT_CANDIDATE
+CPLEX.cpx_callbacksetfunc(model, context_id, my_callback)
+
+CPLEX.optimize!(model)
```
odow: You should see the print `"Aborting!"` and then the solver exit.

examples/admipex9_cb.jl
```diff
@@ -0,0 +1,21 @@
+using CPLEX
+
+env = CPLEX.Env()
+# This is pretty important for thread safety (probably).
+CPLEX.set_param!(env, "CPX_PARAM_THREADS", 1)
```
odow: @chengg04 make sure you set this for now. Once things are wrapped we can have a think about what happens if we let CPLEX use multiple threads...

examples/admipex9_cb.jl
chengg04: Yes, the example does exactly as you said after I updated Julia to 1.0.

chengg04:
@odow I have wrapped all the generaic callback functions following your example from last week. Your code is very clear and helpful. Thank you:)

chengg04:
I plan to write some examples to test the wrapper next weekend. Meanwhile, if anyone has the time to look at the codes and find bugs (I think there are plenty), please let me know! Thanks!

Also, I still get the malloc error 
```
malloc: *** error for object 0x7fa1c2611e80: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
```
for the heuristic callback example admipex9CB_example.jl Line 70 (https://github.com/JuliaOpt/CPLEX.jl/pull/198/commits/72bc238c5e45a4abc5732bc16e2ecd12928ff75a#diff-855ae3c62b5425f62c386446e19db7d8R70). Really get stuck on this and need help...

odow:
Make sure you set `CPLEX.set_param!(env, "CPX_PARAM_THREADS", 1)` in all of the examples.

chengg04:
> Make sure you set `CPLEX.set_param!(env, "CPX_PARAM_THREADS", 1)` in all of the examples.

Added! Thanks for reminding me again.

odow:
So it worked?

chengg04:
No. Still the same error..

chengg04:
I just successfully implemented the generic callback in a context where it's called when integer feasible solution is found (analogous to the lazy callback in the "legacy callback", with some nuances). 

In a hindsight, I should not test on this particular example since it is from a test case for JuMP.jl. I'm going to add some more examples to eliminate bugs.

odow:
What function did you call?

chengg04:
> What function did you call?

```
int CPXPUBLIC
   CPXcallbackpostheursoln (CPXCALLBACKCONTEXTptr context, int cnt,
                            int const *ind, double const *val,
                            double obj,
                            CPXCALLBACKSOLUTIONSTRATEGY strat);
```

And I rewrote it as 
```
function cbpostheursoln(callback_data::CallbackContext,cnt::Int,ind::Vector{Int},val::Vector{Cdouble}, obj::Cdouble, strat::CbSolStrat)

    ind .-= 1

    return_status = @cpx_ccall(callbackpostheursoln,
    Cint,
    (Ptr{Cvoid}, Cint, Ptr{Cint}, Ptr{Cdouble}, Cdouble, Cint),
    callback_data.context, cnt, ind, val, obj, strat)

    # if return_status != 0
    #     throw(CplexError(callback_data.model.env, return_status))
    # end
    # return return_status
    return 0
end
```

PR #210, opened 2019-02-15
compatibility issues

kibaekkim:
I found some lines that returned errors in Julia 1.1.0, and fixed them to make compatible with Julia 1.1.0 and backward too.

### DIFF
--- a/src/cpx_callbacks.jl
+++ b/src/cpx_callbacks.jl
@@ -81,7 +81,7 @@ function cbbranch(cbdata::CallbackData, where::Cint, idx::Cint, LU::Cchar, bd::C
 end
 
 function cbbranchconstr(cbdata::CallbackData, where::Cint, indices::Vector{Cint}, coeffs::Vector{Cdouble}, rhs::Cdouble, sense::Cchar, nodeest::Cdouble)
-    seqnum = Vector{Cint}(1)
+    seqnum = Cint[1]
     stat = @cpx_ccall(branchcallbackbranchconstraints, Cint,
                       (Ptr{Cvoid},
                        Ptr{Cvoid},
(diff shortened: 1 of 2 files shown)


src/cpx_callbacks.jl
```diff
@@ -81,7 +81,7 @@ function cbbranch(cbdata::CallbackData, where::Cint, idx::Cint, LU::Cchar, bd::C
 end
 
 function cbbranchconstr(cbdata::CallbackData, where::Cint, indices::Vector{Cint}, coeffs::Vector{Cdouble}, rhs::Cdouble, sense::Cchar, nodeest::Cdouble)
-    seqnum = Vector{Cint}(1)
+    seqnum = Cint[1]
```
odow: The strict translation would be `Vector{Cint}(undef, 1)`, but your solution also works.

kibaekkim: `Vector{Cint}(undef, 1)` is incompatible with Julia 0.6.4. I was not able to find the corresponding definition in `Compat.jl` either. So, that's my trick :)

mlubin: This trick does not meet readability standards. Someone will look at this code later and wonder why the vector is set to `[1]` when the value of the element is arbitrary. `zeros(CInt, 1)` would be an improvement.
`Vector{Cint}(undef, 1)` should also work if Compat is loaded.

kibaekkim: @mlubin Yes, you are right. I just checked that it works by simply loading `Compat`.

src/cpx_callbacks.jl
```diff
@@ -81,7 +81,7 @@ function cbbranch(cbdata::CallbackData, where::Cint, idx::Cint, LU::Cchar, bd::C
 end
 
 function cbbranchconstr(cbdata::CallbackData, where::Cint, indices::Vector{Cint}, coeffs::Vector{Cdouble}, rhs::Cdouble, sense::Cchar, nodeest::Cdouble)
-    seqnum = Vector{Cint}(1)
+    seqnum = Vector{Cint}(undef,1)
```
blegat: Need a space after the comma to satisfy JuMP style guide

PR #211, opened 2019-02-15
read_model - has_int and has_qc

remi-garcia:
Using read_model and optimize! on MILP get ERROR: CPLEX.CplexError(1017, "CPLEX Error  1017: Not available for mixed-integer problems.\n")
Corrections on MathProgBase.loadproblem! shouldn't be directly on read_model?


odow:
Thanks @remi-garcia

PR #218, opened 2019-02-24
[MOI] Fix handling of parameters

### LINKED ISSUE #216: [Mac OS] No output from CPLEX after running optimize!(mymodel)
The following simple example has been tested but no output is returned.
```julia
julia> using JuMP, CPLEX
julia> mymodel = Model(with_optimizer(CPLEX.Optimizer))
julia> @variable(mymodel, 0 <= x <=2)
julia> @objective(mymodel, Max, x)
julia> optimize!(mymodel)
```
I am sure the model is been properly solved because I can get solution by calling `JuMP.value.(x)`, or `objective_value(mymodel)`.
I tried the same thing on my windows 10 machine and it returned the solving statement correctly.
I tried using another solver GUROBI, `optimize(mymodel, with_optimizer(Gurobi.Optimizer))` and it did returned the solving statement from GUROBI, so I guess there might be something wrong with CPLEX.jl on Mac OS (macOS High Sierra Version 10.13.6 in my case).
I also tried `Pkg.update()` and rebuilt CPLEX by `Pkg.build(CPLEX)` but the issue (no output) still exists.

odow:
Closes #216

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -30,27 +30,45 @@ const SUPPORTED_CONSTRAINTS = [
 
 mutable struct Optimizer <: LQOI.LinQuadOptimizer
     LQOI.@LinQuadOptimizerBase
-    env::Env # Cplex environment
-    Optimizer(::Nothing) = new()
+    env::Union{Nothing, Env}
+    params::Dict{String, Any}
+
+    """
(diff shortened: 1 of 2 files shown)


odow:
Merging this since it's just a direct copy of Gurobi's implementation, it passes all tests, and it fixes the current incorrect behaviour.

PR #224, opened 2019-03-06
fix Base warning

### LINKED ISSUE #223: Deprecated `Base.warn_once`
In https://github.com/JuliaOpt/CPLEX.jl/blob/master/src/CplexSolverInterface.jl#L198

`Base.warn_once` is deprecated, this causes an error in Julia > 0.7

matbesancon:
Fixes #223

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,3 @@
-julia 0.6
+julia 1.0
 MathProgBase 0.5 0.8
-Compat 0.18
 LinQuadOptInterface 0.6 0.7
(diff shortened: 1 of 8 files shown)


odow:
According to REQUIRE, we're still supporting Julia 0.6 so this needs a `Compat`.

Although I think we can remove and bump the minimum Julia version to 1.0 if you like.

matbesancon:
REQUIRE bumped

odow:
Presumably there are some `using Compat`s floating around as well.

odow:
Thanks!

PR #233, opened 2019-04-16
Compute conflict/IIS

dourouc05:
First step for https://github.com/JuliaOpt/JuMP.jl/issues/1035: implement the IIS/conflict within the CPLEX package. 

For now, quite basic: call the conflict refiner, store the information it returns in a structure (allocated only when needed, so that the overhead for people not using it is just one pointer per model).

Next, I will implement MOI attributes that call these functions (if the `conflict` member of `Model` is `nothing`, call the conflict refiner and retrieve its result; if the member is already available, just return the data from it). One question I have for now: I guess the model object may be reused; what happens when it is updated (for instance, the first model is infeasible; the user removes the culprit constraint, then resolves; say they call the conflict refiner again, how is the model reset?).

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -482,6 +482,9 @@ function compute_conflict(model::Optimizer)
         end
     end
     return
+
+    # TODO: decide what to do about the POSSIBLE statuses for the constraints (CPX_CONFLICT_POSSIBLE_MEMBER, 
+    # CPX_CONFLICT_POSSIBLE_UB, CPX_CONFLICT_POSSIBLE_LB). 
 end
 
 function _ensure_conflict_computed(model::Optimizer)
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -547,19 +547,9 @@ function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.Constr
         model.conflict.colstat[var_in_conflict] == CPLEX.CPX_CONFLICT_LB
 end
 
-function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.LE})
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:Union{LQOI.LE, LQOI.GE, LQOI.EQ}})
     _ensure_conflict_computed(model)
-    return (LQOI.cmap(model).less_than[index] - 1) in model.conflict.rowind
-end
-
-function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.GE})
-    _ensure_conflict_computed(model)
-    return (LQOI.cmap(model).greater_than[index] - 1) in model.conflict.rowind
-end
-
-function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.EQ})
-    _ensure_conflict_computed(model)
-    return (LQOI.cmap(model).equal_to[index] - 1) in model.conflict.rowind
+    return (model[index] - 1) in model.conflict.rowind
 end
 
 function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{MOI.ConstraintIndex{<:MOI.SingleVariable, <:Union{LQOI.LE, LQOI.GE}}})
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -456,21 +456,42 @@ function LQOI.make_problem_type_continuous(optimizer::Optimizer)
     set_prob_type!(optimizer.inner, prob_type_toggle_map[prob_type])
     return
 
+"""
+Computes a conflict for the current model. This function should be used in case the model
+is infeasible, in order to determine a subset of the the constraints and of the variable
+that keep the model infeasible (this subset is called a conflict).
+
+Calling this function is required before trying to get any conflict attribute like
+`ConstraintConflictStatus` and `VariableConflictStatus`. Once a conflict is computed
+and the model is modified, the conflict is not automatically purged, and the old conflict
+is returned without a warning.
+"""
 function compute_conflict(model::Optimizer)
     model.conflict = c_api_getconflict(model.inner)
     return
 end
 
+function _ensure_conflict_computed(model::Optimizer)
+    if model.conflict === nothing
+        error("Cannot access conflict status. Call `CPLEX.compute_conflict(model)` first. " *
+            "In case the model is modified, the computed conflict will not be purged.")
+    end
+end
+
 """
     ConflictStatus()
 
-An optimizer attribute indicating the status of the last computed conflict.
+Return an `MOI.TerminationStatusCode` indicating the status of the last computed conflict.
+
+If a minimal conflict is found, it will return `MOI.OPTIMAL`. If the problem is feasible, it will
+return `MOI.INFEASIBLE`. If `compute_conflict` has not been called yet, it will return
+`MOI.OPTIMIZE_NOT_CALLED`.
 """
 struct ConflictStatus <: MOI.AbstractOptimizerAttribute  end
 MOI.is_set_by_optimize(::ConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::ConflictStatus)
-    if model.conflict == nothing
+    if model.conflict === nothing
         return MOI.OPTIMIZE_NOT_CALLED
     elseif model.conflict.stat == CPX_STAT_CONFLICT_MINIMAL
         return MOI.OPTIMAL
@@ -510,6 +531,7 @@ struct ConstraintConflictStatus <: MOI.AbstractConstraintAttribute end
 MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex)
+    _ensure_conflict_computed(model)
     return index in model.conflict.rowind
 end
 
@@ -526,7 +548,8 @@ struct VariableConflictStatus <: MOI.AbstractVariableAttribute end
 MOI.is_set_by_optimize(::VariableConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::VariableConflictStatus, index::MOI.VariableIndex)
-    return in(index, model.conflict.colind)
+    _ensure_conflict_computed(model)
+    return index in model.conflict.colind
 end
 
 function MOI.supports(::Optimizer, ::VariableConflictStatus, ::Type{<:MOI.VariableIndex})
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -490,7 +490,7 @@ If a minimal conflict is found, it will return `MOI.OPTIMAL`. If the problem is 
 return `MOI.INFEASIBLE`. If `compute_conflict` has not been called yet, it will return
 `MOI.OPTIMIZE_NOT_CALLED`.
 """
-struct ConflictStatus <: MOI.AbstractOptimizerAttribute  end
+struct ConflictStatus <: MOI.AbstractModelAttribute  end
 MOI.is_set_by_optimize(::ConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::ConflictStatus)
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -539,17 +539,17 @@ end
 
 function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.LE})
     _ensure_conflict_computed(model)
-    return LQOI.cmap(model).less_than[index] in model.conflict.rowind
+    return (LQOI.cmap(model).less_than[index] - 1) in model.conflict.rowind
 end
 
 function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.GE})
     _ensure_conflict_computed(model)
-    return LQOI.cmap(model).greater_than[index] in model.conflict.rowind
+    return (LQOI.cmap(model).greater_than[index] - 1) in model.conflict.rowind
 end
 
 function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.EQ})
     _ensure_conflict_computed(model)
-    return LQOI.cmap(model).equal_to[index] in model.conflict.rowind
+    return (LQOI.cmap(model).equal_to[index] - 1) in model.conflict.rowind
 end
 
 function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{MOI.ConstraintIndex{<:MOI.SingleVariable, T}}) where {T <: LQOI.LinSets}
@@ -570,7 +570,7 @@ MOI.is_set_by_optimize(::VariableConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::VariableConflictStatus, index::MOI.VariableIndex)
     _ensure_conflict_computed(model)
-    return LQOI.get_column(model, index) in model.conflict.colind
+    return (LQOI.get_column(model, index) - 1) in model.conflict.colind
 end
 
 function MOI.supports(::Optimizer, ::VariableConflictStatus, ::Type{<:MOI.VariableIndex})
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -457,7 +457,7 @@ function LQOI.make_problem_type_continuous(optimizer::Optimizer)
     return
 
 function compute_conflict(model::Optimizer)
-    model.conflict = refineconflict(model.inner)
+    model.conflict = c_api_getconflict(model.inner)
     return
 end
 
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -36,6 +36,7 @@ mutable struct Optimizer <: LQOI.LinQuadOptimizer
     LQOI.@LinQuadOptimizerBase
     env::Union{Nothing, Env}
     params::Dict{String, Any}
+    conflict::Union{Nothing, ConflictRefinerData}
 
     """
         Optimizer(env = nothing; kwargs...)
@@ -69,6 +70,7 @@ function MOI.empty!(model::Optimizer)
     for (name, value) in model.params
         set_param!(model.inner.env, name, value)
     end
+    model.conflict = nothing
     return
 end
 
@@ -453,4 +455,78 @@ function LQOI.make_problem_type_continuous(optimizer::Optimizer)
     # prob_type_toggle_map is defined in file CplexSolverInterface.jl
     set_prob_type!(optimizer.inner, prob_type_toggle_map[prob_type])
     return
+
+function compute_conflict(model::Optimizer)
+    model.conflict = refineconflict(model.inner)
+    return
+end
+
+"""
+    ConflictStatus()
+
+An optimizer attribute indicating the status of the last computed conflict.
+"""
+struct ConflictStatus <: MOI.AbstractOptimizerAttribute  end
+MOI.is_set_by_optimize(::ConflictStatus) = true
+
+function MOI.get(model::Optimizer, attribute::ConflictStatus)
+    if model.conflict == nothing
+        return MOI.OPTIMIZE_NOT_CALLED
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_MINIMAL
+        return MOI.OPTIMAL
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_FEASIBLE
+        return MOI.INFEASIBLE
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_CONTRADICTION
+        return MOI.OTHER_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_DETTIME_LIM
+        return MOI.TIME_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_IT_LIM
+        return MOI.ITERATION_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_MEM_LIM
+        return MOI.MEMORY_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_NODE_LIM
+        return MOI.NODE_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_OBJ_LIM
+        return MOI.OBJECTIVE_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_TIME_LIM
+        return MOI.TIME_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_USER
+        return MOI.OTHER_LIMIT
+    end
+end
+
+function MOI.supports(::Optimizer, ::ConflictStatus)
+    return true
+end
+
+"""
+    ConstraintConflictStatus()
+
+A constraint attribute indicating whether the constraint participates in the last computed conflict.
+"""
+struct ConstraintConflictStatus <: MOI.AbstractConstraintAttribute end
+MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
+
+function MOI.get(model::Optimizer, attribute::ConstraintConflictStatus, index::MOI.ConstraintIndex)
+    return in(index, model.conflict.rowind)
+end
+
+function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{<:MOI.ConstraintIndex})
+    return true
+end
+
+"""
+    VariableConflictStatus()
+
+A variable attribute indicating whether the variable participates in the last computed conflict.
+"""
+struct VariableConflictStatus <: MOI.AbstractVariableAttribute end
+MOI.is_set_by_optimize(::VariableConflictStatus) = true
+
+function MOI.get(model::Optimizer, attribute::VariableConflictStatus, index::MOI.VariableIndex)
+    return in(index, model.conflict.colind)
+end
+
+function MOI.supports(::Optimizer, ::VariableConflictStatus, ::Type{<:MOI.VariableIndex})
+    return true
 end
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -532,7 +532,7 @@ MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex)
     _ensure_conflict_computed(model)
-    return index in model.conflict.rowind
+    return index.value in model.conflict.rowind
 end
 
 function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{<:MOI.ConstraintIndex})
@@ -549,7 +549,7 @@ MOI.is_set_by_optimize(::VariableConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::VariableConflictStatus, index::MOI.VariableIndex)
     _ensure_conflict_computed(model)
-    return index in model.conflict.colind
+    return index.value in model.conflict.colind
 end
 
 function MOI.supports(::Optimizer, ::VariableConflictStatus, ::Type{<:MOI.VariableIndex})
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -564,10 +564,10 @@ function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.Constr
     return (model[index] - 1) in model.conflict.rowind
 end
 
-function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{MOI.ConstraintIndex{<:MOI.SingleVariable, T}}) where {T <: LQOI.LinSets}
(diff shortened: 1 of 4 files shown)


odow:
Why not just call `refineconflict` inside the `c_api_getconflict` function, and return the conflict struct? Given that the model is infeasible, I don't think we need to worry about the performance of unnecessarily calling `refineconflict` multiple times. This also resolves your updating questions: we just call refineconflict every time.

dourouc05:
I got to the MOI attribute step. As suggested in https://github.com/JuliaOpt/JuMP.jl/issues/1035, I tried to be consistent with the existing API: an attribute per constraint and per variable indicating whether they participate in the conflict or not. I could not find an example in MOI where an attribute returns a full-fledged data structure. 

Following the tutorial https://github.com/blegat/JuMP-dev_2019_tutorial/blob/master/Defining%20new%20attributes.ipynb, I defined the attributes as `is_set_by_optimize`: the value of these attributes should not come from the cache, I think. 

To work with that, I added a function to start the conflict refiner (a function that could move to MOI at some point). I feel bad about calling the conflict refiner each time such an attribute is retrieved (that would mean a very large number of calls to the conflict refiner, while only once is enough). 

About the status codes, I used what is already available, even though that looks a bit painful in the eye (I use MOI.INFEASIBLE to indicate that the conflict refiner found no solution because of feasibility…). I could find no mapping for the contradiction status returned by CPLEX (due to the previous use of MOI.INFEASIBLE). Would it be better to add a new enumeration to cover these cases, even when this enum could overlap many termination statuses? 

For now, I completely ignore `rowstat` and `colstat`. I think these should not really be available to the user as such, but rather through the existing interface (one attribute per variable/constraint): these arrays indicate, when the conflict is not minimal, whether a constraint/variable has been proved to be in/out the conflict, or no proof has been found.

src/cpx_model.jl
```diff
@@ -240,3 +240,55 @@ function set_terminate(model::Model)
...
+    stat::Int
+    nrows::Int # number of rows participating in the conflict
+    rowind::Vector{Cint} # index of the rows that participate
+    rowstat::Vector{Cint} # state of the rows that participate
+    ncols::Int # number of columns participating in the conflict
+    colind::Vector{Cint} # index of the columns that participate
+    colstat::Vector{Cint} # state of the columns that participates
+end
+
+function refineconflict(model::Model)
```
odow: Please call this `c_api_getconflict` and ad a comment describing how `refineconflict` is always called first.

src/cpx_model.jl
```diff
@@ -240,3 +240,55 @@ function set_terminate(model::Model)
...
+    # First, compute the conflict. 
+    confnumrows_p = Vector{Cint}(undef, 1)
+    confnumcols_p = Vector{Cint}(undef, 1)
```
odow: To avoid creating a one-element vector, you can use 
```julia
confnumcols_p = Ref{Cint}()
```
Then you can access the value by `confnumcols_p[]`

src/cpx_model.jl
```diff
@@ -240,3 +240,55 @@ function set_terminate(model::Model)
...
+    stat = @cpx_ccall(refineconflict, Cint, (
+                      Ptr{Cvoid},
+                      Ptr{Cvoid},
+                      Ptr{Cint},
+                      Ptr{Cint}
+                      ),
+                      model.env.ptr, model.lp, confnumrows_p, confnumcols_p)
```
odow: I know a lot of the other `ccall`s are indented like this, but we should try moving to something more readable like:
```julia
    stat = @cpx_ccall(
        refineconflict, 
        Cint, 
        (Ptr{Cvoid}, Ptr{Cvoid}, Ptr{Cint}, Ptr{Cint}),
        model.env.ptr, model.lp, confnumrows_p, confnumcols_p)
```

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,79 @@ end
...
 
+"""
+    ConflictStatus()
+
+An optimizer attribute indicating the status of the last computed conflict.
+"""
+struct ConflictStatus <: MOI.AbstractOptimizerAttribute  end
+MOI.is_set_by_optimize(::ConflictStatus) = true
+
+function MOI.get(model::Optimizer, attribute::ConflictStatus)
```
odow: We shouldn't rely on the user calling `compute_conflict`. It is easier if we do something like:
```julia
function MOI.get(model::Optimizer, ::ConflictStatus) 
    if model.conflict === nothing
        model.conflict = c_api_getconflict(model.inner)
    end
    if model.conflict.stat == 
```

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,79 @@ end
...
+"""
+    ConstraintConflictStatus()
+
+A constraint attribute indicating whether the constraint participates in the last computed conflict.
+"""
+struct ConstraintConflictStatus <: MOI.AbstractConstraintAttribute end
+MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
+
+function MOI.get(model::Optimizer, attribute::ConstraintConflictStatus, index::MOI.ConstraintIndex) 
+    return in(index, model.conflict.rowind)
```
odow: This would be clearer as
```julia
function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex) 
    if model.conflict === nothing
        model.conflict = c_api_getconflict(model.inner)
    end
    return index in model.conflict.rowind
end
```

src/MOI_wrapper.jl
odow: We may need to think about checking that the termination status is infeasible, and under what circumstances it's actually okay to call `c_api_getconflict`.

src/MOI_wrapper.jl
odow: Also, no need to name a variable like `attribute` if it isn't used in the function body.

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,79 @@ end
...
+end
+
+function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{<:MOI.ConstraintIndex})
+    return true
+end
+
+"""
+    VariableConflictStatus()
+
+A variable attribute indicating whether the variable participates in the last computed conflict.
```
odow: We should mention that it is a Boolean: "Return a Boolean indicating whether the variable participates in a conflict." Also, we should define "conflict."

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,79 @@ end
...
+"""
+struct VariableConflictStatus <: MOI.AbstractVariableAttribute end
+MOI.is_set_by_optimize(::VariableConflictStatus) = true
+
+function MOI.get(model::Optimizer, attribute::VariableConflictStatus, index::MOI.VariableIndex) 
+    return in(index, model.conflict.colind)
```
odow: See ConstraintConflictStatus

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,79 @@ end
...
+        return MOI.MEMORY_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_NODE_LIM
+        return MOI.NODE_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_OBJ_LIM
+        return MOI.OBJECTIVE_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_TIME_LIM
+        return MOI.TIME_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_USER
+        return MOI.OTHER_LIMIT
+    end
```
odow: There needs to be a final `else` here otherwise the function might return `nothing`.

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,79 @@ end
     end
     # qmatind is ::Vector{Cint}, so we convert back to ::Vector{Int}.
     return sparse(Int.(qmatind), qmatcol, qmatval)
- end
+end
+
+function compute_conflict(model::Optimizer)
+    model.conflict = refineconflict(model.inner)
+    return
+end
```
odow: Probably just remove this function in favour of explicit calls as outlined in the other comments.

dourouc05:
I've addressed the majority of your comments, I think, except for the automatic call to `c_api_getconflict`, as the computed conflict should be wiped out whenever the model changes: your way, the conflict will never be computed again if the model is modified. Or did I miss something?

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,81 @@ end
...
+"""
+    ConstraintConflictStatus()
+
+A Boolean constraint attribute indicating whether the constraint participates in the last computed conflict.
+"""
+struct ConstraintConflictStatus <: MOI.AbstractConstraintAttribute end
+MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
+
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex) 
+    return index in model.conflict.rowind
```
odow: `index` is a `ConstraintIndex`, but `rowind` is `Vector{Cint` so this will always return `false`.

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,81 @@ end
...
+"""
+    ConflictStatus()
+
+An optimizer attribute indicating the status of the last computed conflict.
+"""
+struct ConflictStatus <: MOI.AbstractOptimizerAttribute  end
+MOI.is_set_by_optimize(::ConflictStatus) = true
+
+function MOI.get(model::Optimizer, ::ConflictStatus) 
+    if model.conflict == nothing
```
odow: `model.conflict === nothing`

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,81 @@ end
...
+
+function compute_conflict(model::Optimizer)
+    model.conflict = c_api_getconflict(model.inner)
+    return
+end
+
+"""
+    ConflictStatus()
+
+An optimizer attribute indicating the status of the last computed conflict.
```
odow:
```
"""
Return an `MOI.TerminationStatusCode` indicating the status of the last computed conflict.

If a minimal conflict is found, it will return `MOI.OPTIMAL`. If the problem is feasible, it will
return `MOI.INFEASIBLE`.
"""
```

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,81 @@ end
...
+"""
+    VariableConflictStatus()
+
+A Boolean variable attribute indicating whether the variable participates in the last computed conflict.
+"""
+struct VariableConflictStatus <: MOI.AbstractVariableAttribute end
+MOI.is_set_by_optimize(::VariableConflictStatus) = true
+
+function MOI.get(model::Optimizer, ::VariableConflictStatus, index::MOI.VariableIndex) 
+    return in(index, model.conflict.colind)
```
odow: Prefer `index in model.conflict.colind`, but note that this isn't doing what you think it's doing.

src/cpx_model.jl
```diff
@@ -240,3 +240,49 @@ function set_terminate(model::Model)
...
+
+function c_api_getconflict(model::Model)
+    # This function always calls refineconflict first, which starts the conflict refiner. 
+    # In other words, any call to this function is expensive. 
+
+    # First, compute the conflict. 
+    confnumrows_p = Ref{Cint}()
+    confnumcols_p = Ref{Cint}()
+    stat = @cpx_ccall(
+                refineconflict,
```
odow: The indentation is off. Align 4 spaces in from `stat`.

dourouc05:
> If we're making `compute_conflict` public, it needs a docstring.

Shoudln't we also reset the `conflict` member every time the model is updated, to avoid problems (even though this might have a very slight performance impact)?

odow:
> Shoudln't we also reset the conflict member every time the model is updated

Potentially, but that opens up a can of worms. Let's just force the user to call `compute_conflict` for now.

src/MOI_wrapper.jl
```diff
@@ -489,7 +510,8 @@ struct ConstraintConflictStatus <: MOI.AbstractConstraintAttribute end
 MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex) 
-    return index in model.conflict.rowind
+    _ensure_conflict_computed(model)
+    return index.value in model.conflict.rowind
```
odow: `index.value` isn't guaranteed to be the column. You need:
```julia
LQOI.get_column(model, index) in model.confict.rowind
```

src/MOI_wrapper.jl
```diff
@@ -505,7 +527,8 @@ struct VariableConflictStatus <: MOI.AbstractVariableAttribute end
 MOI.is_set_by_optimize(::VariableConflictStatus) = true
 
 function MOI.get(model::Optimizer, ::VariableConflictStatus, index::MOI.VariableIndex) 
-    return in(index, model.conflict.colind)
+    _ensure_conflict_computed(model)
+    return index.value in model.conflict.colind
```
odow: Ditto on `index.value`

src/MOI_wrapper.jl
```diff
@@ -435,21 +435,42 @@ function LQOI.get_quadratic_terms_objective(model::Optimizer)
...
+"""
 function compute_conflict(model::Optimizer)
     model.conflict = c_api_getconflict(model.inner)
     return
 end
 
+function _ensure_conflict_computed(model::Optimizer)
+    if model.conflict === nothing
+        error("Cannot access conflict status. Call `CPLEX.compute_conflict(model)` first. " * 
+            "In case the model is modified, the computed conflict will not be purged.")
```
odow: Minor thing: can you vertically align the `"` please. So
```julia
        error("Cannot access conflict status. Call `CPLEX.compute_conflict(model)` first. " * 
              "In case the model is modified, the computed conflict will not be purged.")
```

src/MOI_wrapper.jl
```diff
@@ -435,21 +435,42 @@ function LQOI.get_quadratic_terms_objective(model::Optimizer)
     return sparse(Int.(qmatind), qmatcol, qmatval)
 end
 
+"""
+Computes a conflict for the current model. This function should be used in case the model
```
odow: 1) Always begin a docstring with the method. 
2) How about something like:
```julia
"""
    compute_conflict(model::Optimizer)

Compute a minimal subset of the constraints and variables that keep the model
infeasible.

See also `CPLEX.ConflictStatus`, `CPLEX.VariableConflictStatus`, and
`CPLEX.ConstraintConflictStatus`.

Note that if `model` is modified after a call to `compute_conflict`, the
conflict is not purged, and any calls to the above attributes will return values
for the original conflict without a warning.
"""
```

src/MOI_wrapper.jl
dourouc05: Actually, that's the issue I was working on, thanks!

dourouc05: For constraints, however, it does not work as well as expected: I could not find a `get_row` or similar, but rather a `ConstraintMapping` structure (accessible through `cmap`), which is too complicated to work with (there should be a helper method in LQOI to dispatch on the type of constraint). And `index.value` does not seem to be working well enough...

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,107 @@ end
...
+"""
+struct ConstraintConflictStatus <: MOI.AbstractConstraintAttribute end
+MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
+
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex) 
+    _ensure_conflict_computed(model)
+    return LQOI.get_column(model, index) in model.conflict.rowind
+end
+
+function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{<:MOI.ConstraintIndex})
```
odow: So this isn't true for all `MOI.ConstraintIndex`s. It's only for 
```julia
::Type{MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, S} where {S <: LQOI.LinSets}
```
What happens for `VectorAffineFunction` constraints? Or even `ScalarQuadraticFunction` constraints?

What happens for `SingleVariable` constraints (variable bounds)? Should it query the `VariableConflictStatus`?

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,107 @@ end
...
+    ConstraintConflictStatus()
+
+A Boolean constraint attribute indicating whether the constraint participates in the last computed conflict.
+"""
+struct ConstraintConflictStatus <: MOI.AbstractConstraintAttribute end
+MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
+
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex) 
+    _ensure_conflict_computed(model)
+    return LQOI.get_column(model, index) in model.conflict.rowind
```
odow: I think you need something like
```julia
function MOI.get(model::Optimizer, ::ConstraintConflictStatus, 
                 index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.LinSets}) 
    _ensure_conflict_computed(model)
    return model[index] in model.conflict.rowind
end
```

dourouc05: It does not work properly: for the test case I've written, `model[index]` returns `MathOptInterface.VariableIndex(1)`, i.e. the lhs of the constraint, I guess. I've tried to enumerate the cases and write specific methods that use `cmap`, but it's still not working… I'm still investigating other ways to have it work correctly.

odow: That's because your test case uses `SingleVariable` constraints. See my comment https://github.com/JuliaOpt/CPLEX.jl/pull/233#discussion_r277172977. 

The above will only work for `MOI.ScalarAffineFunction` constraints as I put in the function header.

For `MOI.SingleVariable` constraints, you can go
```julia
function MOI.get(model::Optimizer, ::ConstraintConflictStatus, 
                 index::MOI.ConstraintIndex{<:MOI.SingleVariable, <:LQOI.LinSets}) 
    return MOI.get(model, MOI.VariableConflictStatus(), model[index])
end
```

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,139 @@ end
...
+A Boolean constraint attribute indicating whether the constraint participates in the last computed conflict.
+"""
+struct ConstraintConflictStatus <: MOI.AbstractConstraintAttribute end
+MOI.is_set_by_optimize(::ConstraintConflictStatus) = true
+
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.SingleVariable, <:LQOI.LE}) 
+    _ensure_conflict_computed(model)
+    var_in_conflict = findfirst(isequal(LQOI.get_column(model, model[index]) - 1), model.conflict.colind)
+    return model.conflict.colstat[var_in_conflict] == CPLEX.CPX_CONFLICT_MEMBER || 
+        model.conflict.colstat[var_in_conflict] == CPLEX.CPX_CONFLICT_UB
```
odow: What are the other possible values for the conflict?

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,139 @@ end
...
+end
+
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.SingleVariable, <:LQOI.GE})
```
odow: What about `MOI.EqualTo` and `MOI.Interval` constraints?

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,139 @@ end
...
+end
+
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.GE}) 
+    _ensure_conflict_computed(model)
+    return (LQOI.cmap(model).greater_than[index] - 1) in model.conflict.rowind
+end
+
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:LQOI.EQ}) 
+    _ensure_conflict_computed(model)
+    return (LQOI.cmap(model).equal_to[index] - 1) in model.conflict.rowind
```
odow: See my previous comment: https://github.com/JuliaOpt/CPLEX.jl/pull/233/#discussion_r277173014
You should be able to merge these three methods. You shouldn't need to use `cmap`.

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,139 @@ end
...
+end
+
+function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{MOI.ConstraintIndex{<:MOI.SingleVariable, LQOI.LE}})
+    return true
+end
+
+function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{MOI.ConstraintIndex{<:MOI.SingleVariable, LQOI.GE}})
```
odow: Prefer 
```julia
::Type{MOI.ConstraintIndex{<:MOI.SingleVariable, <:Union{LQOI.LE, LQOI.GE}}}
```
over multiple methods.

test/MathOptInterface/MOI_wrapper.jl
```diff
@@ -120,3 +120,25 @@ end
...
 end
+
+@testset "Conflict refiner" begin
+    @testset "Variable bounds" begin
+        # Same test as ../C_API/iis.jl, but ported to MOI. 
+        model = CPLEX.Optimizer()
+        x = MOI.add_variable(model)
+        b = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(0.0))
+        c1 = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(2.0))
+        c2 = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.LessThan(1.0))
```
odow: This isn't the same test. Use `MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([1.0], [x]), 0.0), MOI.LessThan(1.0))` to test scalaraffine constraints.

dourouc05:
While I was working again on this, I wonder whether `VariableConflictStatus` actually makes sense: getting through the documentation of both CPLEX and Gurobi, they both have an attribute for variables, but only indicating whether their upper/lower bounds are included in the conflict/IIS. So I'm thinking of deleting it. What do you think of it? 

(For the rest of your comments, I'm still working on them, focusing more on correctness than niceness for now.)

odow:
If it is only for bounds, then yeah, we probably don't need `VariableConflictStatus`. This is one reason why we're prototyping this in a package first before adding it to MOI.

src/MOI_wrapper.jl
dourouc05: There are six possible values, depending on whether CPLEX could prove a bound is within the conflict or not: 

 * CPLEX.CPX_CONFLICT_MEMBER: both bounds in the conflict
 * CPLEX.CPX_CONFLICT_UB: upper bound in the conflict
 * CPLEX.CPX_CONFLICT_LB: lower bound in the conflict
 * CPLEX.CPX_CONFLICT_POSSIBLE_MEMBER: both bounds POSSIBLY in the conflict
 * CPLEX.CPX_CONFLICT_POSSIBLE_UB: POSSIBLY upper bound in the conflict
 * CPLEX.CPX_CONFLICT_POSSIBLE_LB: POSSIBLY lower bound in the conflict

src/MOI_wrapper.jl
dourouc05: For anything else than affine constraints, another CPLEX function should be called, with more complex things as groups (https://www.ibm.com/support/knowledgecenter/en/SSSA5P_12.5.1/ilog.odms.cplex.help/refcallablelibrary/html/functions/CPXrefineconflictext.html). 

By the way, would you have an example of a VectorAffineFunction constraint?

src/MOI_wrapper.jl
dourouc05: Done in the last commit!

src/MOI_wrapper.jl
dourouc05: They are implemented, I think (and tested also).

dourouc05:
For me, I consider that this is done (I was surprised that affine-in-interval constraints are not supported). Or did I miss something?

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,120 @@ end
...
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:Union{LQOI.LE, LQOI.GE, LQOI.EQ}}) 
+    _ensure_conflict_computed(model)
+    return (model[index] - 1) in model.conflict.rowind
+end
+
+function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{MOI.ConstraintIndex{<:MOI.SingleVariable, T}}) where {T <: LQOI.LinSets}
+    return true
+end
+
+function MOI.supports(::Optimizer, ::ConstraintConflictStatus, ::Type{MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, T}}) where {T <: LQOI.LinSets}
```
odow: Since interval constraints aren't supported, this should just be `T <: Union{LQOI.LE, LQOI.GE, LQOI.EQ}`

Also, you can write
```Julia
function MOI.supports(::Optimizer, ::ConstraintConflictStatus,
                      ::Type{MOI.ConstraintIndex{<:MOI.ScalarAffineFunction, <:Union{LQOI.LE, LQOI.GE, LQOI.EQ}}})
```

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,120 @@ end
...
+    if var_in_conflict === nothing
+        return nothing
+    else
+        return model.conflict.colstat[var_in_conflict]
+    end
+end
 
+function MOI.get(model::Optimizer, ::ConstraintConflictStatus, index::MOI.ConstraintIndex{<:MOI.SingleVariable, <:LQOI.LE}) 
+    status = _sinvar_get_conflict_status(model, index)
+    return status !== nothing && (status == CPLEX.CPX_CONFLICT_MEMBER || status == CPLEX.CPX_CONFLICT_UB)
```
odow: So what to do about the `POSSIBLE` statuses?

dourouc05: They are only generated when `ConflictStatus` is not `MOI.OPTIMAL`. How should we behave in this case? One sensible option would be to discard the distinction between the _POSSIBLE_ and the standard statuses (after all, the user is warned that the result is not optimal). The other option I see is a trilean: true/maybe/false.

test/MathOptInterface/MOI_wrapper.jl
```diff
@@ -120,3 +120,143 @@ end
...
+        x = MOI.add_variable(model)
+        c1 = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.GreaterThan(2.0))
+        c2 = MOI.add_constraint(model, MOI.SingleVariable(x), MOI.LessThan(1.0))
+
+        # Getting the results before the conflict refiner has been called must return an error. 
+        @test MOI.get(model, CPLEX.ConflictStatus()) == MOI.OPTIMIZE_NOT_CALLED
+        @test_throws ErrorException MOI.get(model, CPLEX.ConstraintConflictStatus(), c1)
+
+        # Once it's called, no problem. 
+        CPLEX.compute_conflict(model)
```
odow: Test `ConflictStatus()` after this?

test/MathOptInterface/MOI_wrapper.jl
```diff
@@ -120,3 +120,143 @@ end
...
+        x = MOI.add_variable(model)
+        c1 = MOI.add_constraint(model, MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([1.0], [x]), 0.0), MOI.GreaterThan(2.0))
+        c2 = MOI.add_constraint(model, MOI.ScalarAffineFunction(MOI.ScalarAffineTerm.([1.0], [x]), 0.0), MOI.LessThan(1.0))
+
+        # Getting the results before the conflict refiner has been called must return an error. 
+        @test MOI.get(model, CPLEX.ConflictStatus()) == MOI.OPTIMIZE_NOT_CALLED
+        @test_throws ErrorException MOI.get(model, CPLEX.ConstraintConflictStatus(), c1)
+
+        # Once it's called, no problem. 
+        CPLEX.compute_conflict(model)
```
odow: Ditto on the `ConflictStatus`

dourouc05:
> Is it possible to test a model when the `ConflictStatus` is not `MOI.OPTIMAL` somehow?

That revealed an awkward situation, where `compute_conflict` would error… I added a try-catch to avoid this (while `c_api_getconflict` is as close as possible to CPLEX' behaviour).

src/MOI_wrapper.jl
```diff
@@ -431,5 +433,128 @@ end
...
+for the original conflict without a warning.
+"""
+function compute_conflict(model::Optimizer)
+    try 
+        model.conflict = c_api_getconflict(model.inner)
+    catch exc
+        if isa(exc, CplexError) && exc.code == CPXERR_NO_CONFLICT
+            model.conflict = ConflictRefinerData(CPX_STAT_CONFLICT_FEASIBLE, 0, Cint[], Cint[], 0, Cint[], Cint[])
+        else
+            throw(exc)
```
odow: Use `rethrow(exc)`.

Can you add a comment about why the `try-catch` is necessary?

dourouc05:
Better this way? (There is still a discussion about the POSSIBLE codes that needs to be resolved before merging.)

odow:
Can we test the try-catch please?

src/MOI_wrapper.jl
odow: Maybe we should just add a note to `generate_conflict` and add a TODO about the possible statuses.

dourouc05:
Isn't that what https://github.com/JuliaOpt/CPLEX.jl/pull/233/files#diff-d6eba113d1d7863964be035e6eada3a5R270 is about?

odow:
Oops. I didn't see that test.

dourouc05:
Locally (Windows 10 x64, Julia 1.1.0 x64, CPLEX 12.8.0 x64), they do.

dourouc05:
It seems no one has more comments for this PR; maybe it should be merged?

blegat:
It seems to have conflits. You have to rebase it by `git fetch origin master && git rebase origin/master`

dourouc05:
When I finished working on this PR, there was no conflict at all.

dourouc05:
Here is the rebase. Could this be merged soon?

dourouc05:
@odow @blegat up

odow:
Sorry @dourouc05, on it now.

dourouc05:
Great, thanks!

Issue #234, opened 2019-04-19
Multiple calls to CPLEX.free_problem gives a segfault


trigaut:
@MarcMush notified me that when trying to free many CPLEX problems with the function  `CPLEX.free_problem` we always run into a Segmentation Fault.
https://github.com/JuliaOpt/CPLEX.jl/blob/0ab8e859e05d07f3230d6375b929413eb1d3d291/src/cpx_model.jl#L227

The following example always crashes Julia after 200 or 300 iterations on my computer (Ubuntu 16.04, Julia 1.1, CPLEX1290, Last tagged CPLEX.jl) with a signal(7) : Bus error.

```julia
using JuMP, CPLEX

function build_direct_model()
    m = direct_model(CPLEX.Optimizer(CPX_PARAM_SCRIND=0))
    @variable(m, 0 <= x <= 1)
    @objective(m, Min, x)
    return m
end


function run_direct_models()
	k=0

	while true
		k+=1
		println(k)
		m = build_direct_model()
		CPLEX.free_problem(m.moi_backend.inner)
		sleep(0.1)
	end
end 

run_direct_models()
```
Same issue with caching JuMP models instead of direct models (after a call to `optimize!`).

The following fix seems to solve the issue and still free the memory.

```julia
function free_problem(model::Model)
    tmp = model.lp # instead of Ptr{Cvoid}[model.lp]
    stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, tmp)
    if stat != 0
        throw(CplexError(model.env, stat))
    end
end
```

Or did we get something wrong?

odow:
Your fix seems correct. The `Ptr{Cvoid}[model.lp]` is just an artefact from 5 years ago: https://github.com/JuliaOpt/CPLEX.jl/commit/4d8f78fabd36daea6e8ec6ca6317258d457925b8

odow:
Closed by #235

Issue #27, opened 2014-09-26
Cplex environments are not freed


ghost:
Each time `Env()` is called a new cplex environment is created using `CPXopenCplex`. Unfortunately, no finalizer is registered, so creating a lot of models as in the following example eventually eats up your memory:

``` julia
using MathProgBase
using CPLEX
for i=1:10000
  local m = MathProgBase.model(CplexSolver())
end
```

I suggest to add a corresponding finalizer:

``` diff
diff --git a/src/cpx_env.jl b/src/cpx_env.jl
index af7f19e..6c58e69 100644
--- a/src/cpx_env.jl
+++ b/src/cpx_env.jl
@@ -2,12 +2,14 @@ type Env
     ptr::Ptr{Void}

     function Env()
-      stat = Array(Cint, 1)
-      tmp = @cpx_ccall(openCPLEX, Ptr{Void}, (Ptr{Cint},), stat)
-      if tmp == C_NULL
-          error("CPLEX: Error creating environment")
-      end
-      new(tmp)
+        stat = Array(Cint, 1)
+        tmp = @cpx_ccall(openCPLEX, Ptr{Void}, (Ptr{Cint},), stat)
+        if tmp == C_NULL
+            error("CPLEX: Error creating environment")
+        end
+        env = new(tmp)
+        finalizer(env, close_CPLEX)
+        env
     end
 end
```

The two functions `free_problem` and `close_CPLEX` appear to have a bug: the argument of the object to be closed must be a **pointer to** `CPXLPptr` and `CPXENVptr`, respectively.

``` diff
diff --git a/src/cpx_model.jl b/src/cpx_model.jl
index ba53f3b..2aecb99 100644
--- a/src/cpx_model.jl
+++ b/src/cpx_model.jl
@@ -136,15 +136,17 @@ function set_warm_start!(model::Model, indx::IVec, val::FVec)
 end

 function free_problem(model::Model)
-    stat = @cpx_ccall(freeprob, Cint, (Ptr{Void}, Ptr{Void}), model.env.ptr, model.lp)
+    tmp = Ptr{Void}[model.lp]
+    stat = @cpx_ccall(freeprob, Cint, (Ptr{Void}, Ptr{Void}), model.env.ptr, tmp)
     if stat != 0
         throw(CplexError(model.env, stat))
     end
 end

 function close_CPLEX(env::Env)
-    stat = @cpx_ccall(closeCPLEX, Cint, (Ptr{Void},), env.ptr)
[...]

PR #235, opened 2019-04-19
fix free_problem

MarcMush:
solve #234

### DIFF
--- a/src/cpx_model.jl
+++ b/src/cpx_model.jl
@@ -225,8 +225,7 @@ function set_warm_start!(model::Model, indx::IVec, val::FVec, effortlevel::Integ
 end
 
 function free_problem(model::Model)
-    tmp = model.lp
-    stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, tmp)
+    stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, model.lp)
     if stat != 0
         throw(CplexError(model.env, stat))
     end
--- a/src/cpx_model.jl
+++ b/src/cpx_model.jl
@@ -225,7 +225,7 @@ function set_warm_start!(model::Model, indx::IVec, val::FVec, effortlevel::Integ
 end
 
 function free_problem(model::Model)
-    tmp = Ptr{Cvoid}[model.lp]
+    tmp = model.lp
     stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, tmp)
     if stat != 0
         throw(CplexError(model.env, stat))
--- a/src/cpx_model.jl
+++ b/src/cpx_model.jl
@@ -11,7 +11,11 @@ function Model(env::Env, lp::Ptr{Cvoid})
     notify_new_model(env)
     model = Model(env, lp, false, false, false, nothing, Cint[0])
     function model_finalizer(model)
-        free_problem(model)
+        if model.lp != C_NULL
+            free_problem(model)
+        else
+            # User must have called `free_problem` directly.
+        end
         notify_freed_model(env)
     end
     finalizer(model_finalizer, model)
@@ -226,6 +230,7 @@ end
 
 function free_problem(model::Model)
     stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, model.lp)
+    model.lp = C_NULL
     if stat != 0
         throw(CplexError(model.env, stat))
     end


src/cpx_model.jl
```diff
@@ -225,7 +225,7 @@ function set_warm_start!(model::Model, indx::IVec, val::FVec, effortlevel::Integ
 end
 
 function free_problem(model::Model)
-    tmp = Ptr{Cvoid}[model.lp]
+    tmp = model.lp
     stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, tmp)
```
odow: Please in-line `tmp`, so that it is just
```julia
stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, model.lp)
```

Also, can you run the tests `] test CPLEX` and confirm that they still all pass.

MarcMush:
in-line done
all tests still pass

odow:
Okay, so I dug into this.

From the [CPLEX docs for CPXfreeProb](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.9.0/ilog.odms.cplex.help/refcallablelibrary/cpxapi/freeprob.html), `lp_p` is `A pointer to a CPLEX problem object as returned by CPXXcreateprob/CPXcreateprob.`, not a variable holding a pointer [as originally thought](https://github.com/JuliaOpt/CPLEX.jl/issues/27#issue-44026538).

The confusion stems from the fact that `closeCPLEX` requires [`A pointer to a variable holding the pointer to the CPLEX environment as returned by CPXXopenCPLEX/CPXopenCPLEX.`](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.9.0/ilog.odms.cplex.help/refcallablelibrary/cpxapi/closeCPLEX.html). So we have
https://github.com/JuliaOpt/CPLEX.jl/blob/0ab8e859e05d07f3230d6375b929413eb1d3d291/src/cpx_env.jl#L44-L51

mlubin:
Is `model.lp` set to `C_NULL` after the call? If not, we should do so to avoid crashes.

odow:
> Is model.lp set to C_NULL after the call?

No. Good point.

MarcMush:
> Is `model.lp` set to `C_NULL` after the call?

It throws an error when `model.lp` is set to `C_NULL` after the call (usually after 600 calls of `CPLEX.free_problem`) :
```
error in running finalizer: CPLEX.CplexError(code=1004, msg="CPLEX Error  1004: Null pointer for required data.
")
```

trigaut:
> > Is `model.lp` set to `C_NULL` after the call?
> 
> It throws an error when `model.lp` is set to `C_NULL` after the call (usually after 600 calls of `CPLEX.free_problem`) :
> 
> ```
> error in running finalizer: CPLEX.CplexError(code=1004, msg="CPLEX Error  1004: Null pointer for required data.
> ")
> ```

I confirm with the script of Issue #234 .

I tried to add `notify_freed_model(model.env)` such as:
```
function free_problem(model::Model)
    stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, model.lp)
    model.lp = C_NULL
    notify_freed_model(model.env)
    if stat != 0
        throw(CplexError(model.env, stat))
    end
end
```

but I obtain 
`error in running finalizer: AssertionError(msg="env.num_models > 0")`
after around 500 iterations of script of Issue #234 .

mlubin:
Clearly we should avoid making `cpx_ccall`s when either the environment or lp model pointer is `C_NULL`.

trigaut:
Just like this?

```julia
function free_problem(model::Model)
  if (model.env != C_NULL) && (model.lp != C_NULL)
    stat = @cpx_ccall(freeprob, Cint, (Ptr{Cvoid}, Ptr{Cvoid}), model.env.ptr, model.lp)
    model.lp = C_NULL
    if stat != 0
        throw(CplexError(model.env, stat))
    end
  end
end
```
It seems to work.
We should not use `notify_freed_model(model.env)`?

odow:
Instead of `free_prolem`, modify
https://github.com/JuliaOpt/CPLEX.jl/blob/0ab8e859e05d07f3230d6375b929413eb1d3d291/src/cpx_model.jl#L13-L16
to be
```julia
    function model_finalizer(model)
        if model.lp != C_NULL
            free_problem(model)
        else
            # User must have called `free_problem` directly.
        end
        notify_freed_model(env)
    end
```

odow:
Almost there @MarcMush. Can you confirm that the tests still pass and that the script in #234 runs without crashing.

trigaut:
I tested @MarcMush  branch. 
Everything passes and the script in #234 does not crash.
testscplex

odow:
Thanks @MarcMush and @trigaut

PR #238, opened 2019-05-10
add function to switch cplex prob type

vitornesello:
Related to this issue: https://github.com/atoptima/Coluna.jl/issues/56

I tried to use CPLEX as LP/MILP solver in Coluna as follows:
Use it as LP solver during the column generation iterations so solve the master LP and then at the end of the node switch to MILP to solve the IP version of the restricted master.

In order to change from one to the other I remove/ add all single variable integrality constraints.

What happens is that when I start to solve the second node (after removing all the integrality constraints), CPLEX is still in MILP mode, so I cannot get my dual values needed for the column generation.

I went to the code and saw that the function that changes the Cplex problem type (called if you just removed all integrality constraint or added one integrality constraint) was not implemented. There was a 'dirty fix' in place.
So I implemented in both senses (MILP -> LP and LP->MILP) and it worked just fine.

### DIFF
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -180,12 +180,15 @@ function LQOI.delete_variables!(model::Optimizer,
 end
 
 function LQOI.solve_mip_problem!(model::Optimizer)
-    model.inner.has_int = true
-    LQOI.solve_linear_problem!(model)
+    LQOI.make_problem_type_integer(model)
+    optimize!(model.inner)
+    return
 end
 
-function LQOI.solve_linear_problem!(model::Optimizer)  
+function LQOI.solve_linear_problem!(model::Optimizer)
+    LQOI.make_problem_type_continuous(model)
     optimize!(model.inner)
+    return
 end
 
 function LQOI.get_termination_status(model::Optimizer)
@@ -267,4 +270,16 @@ end
 
 function LQOI.get_objective_value(model::Optimizer)
     return c_api_getobjval(model.inner)
+end
+
+function LQOI.make_problem_type_integer(optimizer::Optimizer)
+    optimizer.inner.has_int = true
+    set_prob_type!(optimizer.inner, :MILP)
+    return
+end
+
+function LQOI.make_problem_type_continuous(optimizer::Optimizer)
+    optimizer.inner.has_int = false
+    set_prob_type!(optimizer.inner, :LP)
+    return
 end


src/MOI_wrapper.jl
```diff
@@ -433,3 +434,14 @@ end
     return sparse(Int.(qmatind), qmatcol, qmatval)
  end
 
+function LQOI.make_problem_type_integer(optimizer::Optimizer)
+    optimizer.inner.has_int = true
+    set_prob_type!(optimizer.inner, :MILP)
```
odow: You need to check for quadratics as well. Check the MPB code. I think there are helper functions.

vitornesello: Ok. Tests are all screwed-up for now

vitornesello: and by chance they fail only for non-linear stuff

vitornesello:
Now tests pass for all problem types. I think it would be good to add a test that tests when you switch back and forth integer/linear.

src/MOI_wrapper.jl
```diff
@@ -432,16 +432,23 @@ end
     end
     # qmatind is ::Vector{Cint}, so we convert back to ::Vector{Int}.
     return sparse(Int.(qmatind), qmatcol, qmatval)
- end
+end
+
+const integer_types = Set{Symbol}([:MILP, :MIQP, :MIQCP])
+const continuous_types = Set{Symbol}([:LP, :QP, :QCP])
```
odow: Constants should be UPPER_CASED.

src/MOI_wrapper.jl
```diff
@@ -432,16 +432,23 @@ end
...
 
 function LQOI.make_problem_type_integer(optimizer::Optimizer)
     optimizer.inner.has_int = true
-    set_prob_type!(optimizer.inner, :MILP)
+    prob_type = get_prob_type(optimizer.inner)
+    prob_type in integer_types && return
+    set_prob_type!(optimizer.inner, prob_type_toggle_map[prob_type])
```
odow: Add a comment that `prob_type_toggle_map` is defined in CplexSolverInterface.jl.

src/MOI_wrapper.jl
```diff
@@ -432,16 +432,23 @@ end
...
     return
 end
 
 function LQOI.make_problem_type_continuous(optimizer::Optimizer)
     optimizer.inner.has_int = false
-    set_prob_type!(optimizer.inner, :LP)
+    prob_type = get_prob_type(optimizer.inner)
+    prob_type in continuous_types && return
+    set_prob_type!(optimizer.inner, prob_type_toggle_map[prob_type])
```
odow: Ditto comment.

test/runtests.jl
```diff
@@ -3,8 +3,11 @@ import Pkg
...
-        include(joinpath(folder, file))
-    end
+@testset "MOI" begin
+    include("MathOptInterface/MOI_wrapper.jl")
 end
+# @testset "$folder" for folder in ["C_API", "MathProgBase", "MathOptInterface"]
+#     @testset "$(file)" for file in readdir(folder)
+#         include(joinpath(folder, file))
+#     end
+# end
```
odow: Please revert this once you've finished testing.

odow:
> I think it would be good to add a test that tests when you switch back and forth integer/linear.

Good idea. Just add it to the CPLEX tests for now.

vitornesello:
Now I can confirm that tests pass on Mac using Julia 1.0 and 1.1.

vitornesello:
Guys is this good to merge?

odow:
Thanks

PR #244, opened 2019-07-08
Fix syntax error in LQOI.make_problem_type_integer.

tkoolen:


odow:
Introduced by https://github.com/JuliaOpt/CPLEX.jl/pull/233. Another reason for needing automated tests...

PR #248, opened 2019-08-09
Check for CPLEX in the PATH environment variable

dourouc05:
This should ease installation in some cases.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -40,16 +40,16 @@ base_cpxvers = ["128", "129"]
 cpxvers = [base_cpxvers; base_cpxvers .* "0"]
 
 libnames = String["cplex"]
-for v in reverse(cpxvers)
-    if Sys.isapple()
-        push!(libnames, "libcplex$v.dylib")
-    elseif Sys.isunix()
-        push!(libnames, "libcplex$v.so")
+
+@static if ! Sys.iswindows()
+    extension = if Sys.isapple(); ".dylib"; else; ".so"; end
+    for v in reverse(cpxvers)
+        push!(libnames, "libcplex$(v).$(extension)")
         if haskey(ENV, base_env)
-            push!(libnames, joinpath(ENV[base_env], "libcplex$v.so"))
+            push!(libnames, joinpath(ENV[base_env], "libcplex$(v).$(extension)"))
         end
         if cplex_path !== nothing
-            push!(libnames, joinpath(cplex_path, "libcplex$v.so"))
+            push!(libnames, joinpath(cplex_path, "libcplex$(v).$(extension)"))
         end
     end
 end
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -36,7 +36,8 @@ if cplex_path !== nothing
     cplex_path = dirname(strip(cplex_path))
 end
 
-const cpxvers = ["128", "1280", "129", "1290"]
+base_cpxvers = ["128", "129"]
+cpxvers = [base_cpxvers; base_cpxvers .* "0"]
 
 libnames = String["cplex"]
 for v in reverse(cpxvers)
@@ -53,23 +54,20 @@ for v in reverse(cpxvers)
     end
 end
 
-const wincpxvers = ["128", "1280", "129", "1290"]
+base_wincpxvers = ["128", "129"]
+wincpxvers = [base_wincpxvers; base_wincpxvers .* "0"]
+
 @static if Sys.iswindows()
     for v in reverse(wincpxvers)
         env = base_env * v
         if haskey(ENV, env)
             for d in split(ENV[env], ';')
                 occursin("cplex", d) || continue
-                if length(v) == 3 # annoying inconsistency
-                    push!(libnames, joinpath(d, "cplex$(v)0"))
-                else
-                    push!(libnames, joinpath(d, "cplex$(v)"))
-                end
+                push!(libnames, joinpath(d, "cplex$(v)"))
             end
         end
         if cplex_path !== nothing
             push!(libnames, joinpath(cplex_path, "cplex$(v)"))
-            push!(libnames, joinpath(cplex_path, "cplex$(v)0"))
         end
     end
 end
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -20,6 +20,11 @@ end
 end
 
 base_env = "CPLEX_STUDIO_BINARIES"
+cplex_path = splitdir(strip(try
+        read(`which cplex`, String)
+    catch
+        nothing
+    end))[1]
 
 const cpxvers = ["128", "1280", "129", "1290"]
 
@@ -31,6 +36,9 @@ for v in reverse(cpxvers)
         push!(libnames, "libcplex$v.so")
         if haskey(ENV, base_env)
             push!(libnames, joinpath(ENV[base_env], "libcplex$v.so"))
+        end
+        if cplex_path !== nothing
+            push!(libnames, joinpath(cplex_path, "libcplex$v.so"))
         end
     end
 end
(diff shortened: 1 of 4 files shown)


dourouc05:
I used the "new" functions `Sys.isapple()` and the like: as CPLEX.jl does not support Julia 0.7- (the `Project.toml` file seems to imply that), maybe it's time to clean up the build file to remove anything from before Julia 1.0?

deps/build.jl
```diff
@@ -9,46 +9,66 @@ end
...
-        print(f,"const libcplex = ")
+        print(f, "const libcplex = ")
         show(f, path) # print with backslashes excaped on windows
         println(f)
     end
 end
 
-@static if (VERSION >= v"0.7.0-DEV.3382" && Sys.isapple()) || (VERSION < v"0.7.0-DEV.3382" && is_apple())
-    Libdl.dlopen("libstdc++",Libdl.RTLD_GLOBAL)
+@static if Sys.isapple()
```
blegat: Thanks for the cleanup!

deps/build.jl
```diff
@@ -9,46 +9,66 @@ end
...
 
 base_env = "CPLEX_STUDIO_BINARIES"
 
+cplex_path = try
+        @static if Sys.isapple() || Sys.isunix()
+            read(`which cplex`, String)
+        elseif Sys.iswindows()
+            read(`where cplex`, String)
+        end
+    catch
```
blegat: The indendation of this block should be decreased by 1 IMO as `try` is in an unindented block

deps/build.jl
```diff
@@ -9,46 +9,66 @@ end
...
+        nothing
+    end
+if cplex_path !== nothing
+    cplex_path = splitdir(strip(cplex_path))[1]
```
blegat: Could you add a comment on usually what's the value of `cplex_path` after and before this line ?

deps/build.jl
dourouc05: Shall I make a PR with similar changes throughout the package? (The question being: is there a rewrite effort in progress, like for Gurobi and GLPK?)

deps/build.jl
```diff
@@ -2,53 +2,75 @@
...
+        read(`which cplex`, String)
+    elseif Sys.iswindows()
+        read(`where cplex`, String)
+    end
+catch
+    nothing
+end
+if cplex_path !== nothing
+    # Extract the path to the folder containing the CPLEX executable.
+    cplex_path = splitdir(strip(cplex_path))[1]
```
blegat: Isn't `splitdir(s)[1]` equivalent to `dirname(s)` ?

deps/build.jl
```diff
@@ -2,53 +2,75 @@
...
+                    push!(libnames, joinpath(d, "cplex$(v)0"))
                 else
-                    push!(libnames,joinpath(d,"cplex$(v)"))
+                    push!(libnames, joinpath(d, "cplex$(v)"))
                 end
             end
         end
+        if cplex_path !== nothing
+            push!(libnames, joinpath(cplex_path, "cplex$(v)"))
+            push!(libnames, joinpath(cplex_path, "cplex$(v)0"))
```
blegat: Why adding the `0` ? It is already added in the `wincpxvers` array, isn't it ?

dourouc05: Actually, I duplicated what is done above, when checking with the environment variable.

blegat: Yes, but to be sure that zeros are in the `wincpxvers` array, I would do
```julia
base_wincpxvers = ["128", "129"]
wincpxvers = [base_wincpxvers; base_wincpxvers .* "0"]
```
and same for `cpxvers` for consistency

dourouc05: I just did this. How about having only one array instead of two (wincpxvers and cpxvers)?

blegat: No idea why there are two, they used to be different but I don't know whether we might need two different list again in the future.

dourouc05:
As this PR is approved, could it get merged?

deps/build.jl
```diff
@@ -2,53 +2,73 @@
...
         push!(libnames, "libcplex$v.dylib")
-    elseif (VERSION >= v"0.7.0-DEV.3382" && Sys.isunix()) || (VERSION < v"0.7.0-DEV.3382" && is_unix())
+    elseif Sys.isunix()
         push!(libnames, "libcplex$v.so")
         if haskey(ENV, base_env)
             push!(libnames, joinpath(ENV[base_env], "libcplex$v.so"))
         end
+        if cplex_path !== nothing
+            push!(libnames, joinpath(cplex_path, "libcplex$v.so"))
+        end
```
odow: Why only add this is `isiunix`? What about the `is apple` and the `.dylib`?

deps/build.jl
```diff
@@ -2,53 +2,73 @@
...
     end
 end
 
-const wincpxvers = ["128", "1280", "129", "1290"]
-@static if (VERSION >= v"0.7.0-DEV.3382" && Sys.iswindows()) || (VERSION < v"0.7.0-DEV.3382" && is_windows())
+base_wincpxvers = ["128", "129"]
+wincpxvers = [base_wincpxvers; base_wincpxvers .* "0"]
```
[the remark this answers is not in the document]
dourouc05: That question has already been raised. I suppose your comment means that one of them can leave.

deps/build.jl
dourouc05: I did something in the latest commit, but I have no possibility to test on macOS… Does that look right to you? (Maybe I should merge the two `@static if`?)

odow: Yeah, can't this whole file just be tidied up and simplified?

dourouc05:
I've just simplified the code quite a bit by merging all code paths for the different OSes. 

I really like what you did for Gurobi, the diagnoses. Should a similar implementation go into this PR or rather another one? 

This also points to a discrepancy between the two packages: for CPLEX.jl, just finding the CPLEX executable on the path is enough; for Gurobi.jl, the environment variable must be set. I think that using the version on the path is a very reasonable default (and should be implemented in Gurobi.jl). To be sure that the Julia package finds the version the user really wants, maybe we should print it when building the Julia package?

dourouc05:
I just pushed a few more changes, mostly cleaning things up. I also tried to extract the version within `build.jl`, which could also be a nice sanity check, but it's a nightmare to make it work, as you cannot just include the relevant functions from the rest of the package (I start getting "world age" errors). Here is what I was writing (the `eval`s avoid errors about not giving a Ptr, but a Tuple{String, String}, even though that's what all CPLEX.jl is doing… I guess this is a Julia bug); I was stuck at typing issues for `stat` (), but I'm giving up right now: 

```
function get_version(path)
    # Cannot use CPLEX.version
    version = "nothing"

    if Sys.iswindows()
        # Create environment.
        stat = Vector{Cint}(undef, 1)
        env = eval(:(ccall(("CPXopenCPLEX", path), stdcall, Ptr{Cvoid}, (Ptr{Cint},), stat)))

        # Get version.
        version_char_ptr = eval(:(ccall(("version", path), stdcall, Ptr{Cchar}, (Ptr{Cvoid},), env)))
        if version_char_ptr == C_NULL
            error("CPLEX: error getting version")
        end
        version = unsafe_string(version_char_ptr)

        # Close CPLEX.
        eval(:(ccall(("CPXcloseCPLEX", path), stdcall, Cint, (Ptr{Cvoid},), env)))
    end

    return version
end
```

This works locally (Windows 10 1709, Julia 1.2 RC2): 

```
(v1.2) pkg> build CPLEX
  Building CPLEX → `C:\Users\XXX\.julia\dev\CPLEX\deps\build.log`
 Resolving package versions...
```

odow:
You could add a print with `Using CPLEX found in location $(libname)` if you like.

dourouc05:
I just did that in the last commit. The only problem is that this is printed out in the log file, but I'm not sure anything else is possible.

PR #256, opened 2019-10-16
Update README.md

salazarzen:


blegat:
Thanks this is indeed needed for Julia v1.0 and later.

PR #258, opened 2019-10-29
Test CPLEX on travis

odow:
I'm embarking on a CPLEX.jl re-write.

Step 1: get CPLEX running on Travis
Step 2: write the MOI wrapper
Step 3: replace the existing C wrapper with one generated by Clang.jl
Step 4: change MOI wrapper as needed to account for the Clang wrapper.

The `libcplex.so` is currently stored in a public storage provider accessible with the correct URL, which is stored in a secured environment variable:
https://docs.travis-ci.com/user/environment-variables#defining-variables-in-repository-settings

By default, Pull Requests that come from a fork of this repo can't access the `SECRET_URL_KEY` so they will fail: https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions

Thus, only JuliaOpt members with commit access can trigger builds using the CPLEX binary. This isn't a good long-term solution, but it can be ameliorated by us distributing the community edition, and using that on Travis.

cc @mlubin

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ branches:
   - "/^release-.*$/"
 env:
   global:
-    secure: DlLHJfhnRTodFqT1vbxkX5qwaSPKX5geq09kewkmk4TE1COFMeJXfH2KzpXCpKy0BREYuk5ZCDp5zNsYspNKzRt3Libl9Utz948NKcYSDgKRPAfvqs8Y5R57/TQvcvq6Y5UYPQ4USnktfEuXLCmJeGrnp1WtXUpvD5h6RjHgeKk=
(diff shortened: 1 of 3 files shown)


mlubin:
The restriction on running CI on pull requests is pretty significant, but better to have some CI than none.

odow:
> The restriction on running CI on pull requests is pretty significant

Yeah. But since users can execute arbitrary code through a PR, I don't see any other options. It'd be nice for us to be able to manually kick-off a run if we've reviewed the code and it isn't doing anything dodgy.

jd-lara:
I would suggest changing the order of the operations and getting the code from Clang first. The experience so far with Xpress is that is better to be able to customize the Clang code as needed by MOI.

odow:
> I would suggest changing the order of the operations and getting the code from Clang first.

I considered this, but it should be easy to copy the Gurobi wrapper first, and then make changes to the underlying API. I'll add Step 4: change MOI wrapper as needed to account for the Clang wrapper.

kdheepak:
I'd be interested in seeing if you have any neat tricks to make the wrapper generated by Clang.jl safe to use. We are doing the same thing for Xpress.jl (https://github.com/JuliaOpt/Xpress.jl/pull/43) and we are settling on some ideas, but it'd be discuss it with you as well to see if you have other suggestions.

odow:
Merging this so that I have CI going forward.

PR #250, opened 2019-08-23
WIP: New MOIWrapper

jd-lara:
Work in progress for https://github.com/JuliaOpt/MathOptInterface.jl/issues/736

PR #261, opened 2019-10-31
WIP: new CPLEX wrapper

### LINKED ISSUE #98: warn when using callbacks and threads
https://github.com/JuliaOpt/JuMP.jl/issues/904

### LINKED ISSUE #200: Not really an issue, delete a constraint that genereate conflict
I'm sorry to open an issue but I didn't find a blog in which I can ask the following question, please close this issue as soon as you prefer.

I would like to delete a constraint from an infeasible model, in particular when I run `solve(m)` I have an infeasibility generated by, let's say, constraint 'c2000', how can I delete this constraint using this CPLEX internal index?

Thank you  for your attention.
Giada

### LINKED ISSUE #222: How to add callbacks?
Hi,
I am a bit puzzled on who is "responsible" for my question, hence, I opened a thread also [here](https://discourse.julialang.org/t/adding-cplex-callbacks-with-jump-v0-19/21449). In a nutshell, I updated to JuMP v0.19, and it's documentation says that callbacks are now solver-specific. However, I do not find something about adding callbacks here in the documentation nor in the MathOptInterface docs. 
So my question is, what is the right way to add callbacks?
Thank you in advance.

### LINKED ISSUE #225: Pass constraint names to CPLEX
Hi,

All CPLEX solver messages are reported with generic constraint names e.g. c23474

It makes it very difficult to diagnose infeasibilities and such.

Is there a plan to pass the JuMP constraint names on to CPLEX? It seems possible as in GAMS for example I get more helpful names.

### LINKED ISSUE #230: Gap in JuMP v0.19
Hello,
When I try to obtain in the MIP the relative gap and number of nodes, it gave me error:
```julia
gap_info = MOI.get(model, MOI.RelativeGap())
Num_nodes = MOI.get(model, MOI.NodeCount())
```
The error message is
```
MethodError: no method matching get_relative_mip_gap(::CPLEX.Optimizer)
Closest candidates are:
get_relative_mip_gap(!Matched::LinQuadOptInterface.MockLinQuadOptimizer) at
```
However, I used the same code to retrieve ObjectiveBound() and CPU time it works. It seems like some of the attributes are not working. Could you give me some insight about why would this happen or to implement in CPLEX.jl in a new version? 
Thank you.

### LINKED ISSUE #241: Add support for Semicontinuous
It might be as simple as adding a line here:
https://github.com/JuliaOpt/CPLEX.jl/blob/master/src/MOI_wrapper.jl#L21
similar to
https://github.com/JuliaOpt/Gurobi.jl/blob/master/src/MOI_wrapper.jl#L28-L29

### LINKED ISSUE #242: Binary Variables with Bounds
As discussed in this related issue, https://github.com/lanl-ansi/PowerModels.jl/issues/535

A code like the following (untested) appears to result in an error,
```
using JuMP; using CPLEX
m = Model()
@variable(m, x, binary=true, lower_bound=0, upper_bound=1)
optimize!(m, with_optimizer(CPLEX.Optimizer))
```

### LINKED ISSUE #243: Support VariablePrimalStart
https://discourse.julialang.org/t/variableprimalstart-is-not-supported-by-the-model-and-lazybridgeoptimizer/25447

### LINKED ISSUE #249: Update wrapper to MOI 0.9
follow the gurobi implementation: JuliaOpt/Gurobi.jl#216

link the PR to JuliaOpt/MathOptInterface.jl#736

cc @jd-lara

odow:
Nominally passing tests with exclusions/hacks. Remaining TODO items:

- [ ] delete SOC variables
- [x] number threads
- [x] solve time
- [ ] duals for quadratic problems
- [ ] duals for SOC constraints
- [x] set constraint names in C
- [x] set variable names in C
- [x] ~pass model name to inner~
- [x] Callbacks
- [x] RawStatusString
- [x] PrimalStatus check for unbounded ray
- [x] SimplexIterations
- [x] BarrierIterations
- [x] VariablePrimalStart
- [ ] Basis status
- [x] Constraint conflicts

At a minimum, this PR:
Closes #98
Closes #200
Closes #222
Closes #225
Closes #230
Closes #241
Closes #242
Closes #243
Closes #249
Closes #250


odow:
I've added MOI callbacks. 

Note: This is a breaking change for people using the existing generic CPLEX callbacks since there were a few bugs. I've commented some out and changed the inputs to be 0-indexed using `Cint` instead of 1-indexed with `Int`.

However, I don't think this is an issue given that I think we should re-write the existing C -> Julia wrapper with Clang. The next release is going to be massively breaking anyway.

odow:
I'm in favor of merging this as-is, and opening issues to address the remaining TODO's.

odow:
I'm guessing no one will want to/have time to review this, so I will leave this open another day and then merge. The tests passing CI (thanks IBM!) and the number of closed issues, speak for themselves.

joaquimg:
@odow how many hours did you expend getting this up and running?

odow:
It was only a day and a couple of odd bits and pieces. It's pretty similar to Gurobi. I think Xpress is a different story. Although check the solution caching. I think you need to do something similar?

joaquimg:
CPLEX might be even closer to xpress. I will take a look.

PR #273, opened 2020-01-03
Specialization of MOI.delete for vector of variables

henriquebecker91:
Arised from a discussion with @odow and @mlubin.

MOI.delete already has a fallback method for a vector of variable parameters (which just calls the single variable delete over each variable of the vector). In Gurobi, the problem is that this fallback has O(n^2) performance the way the things are now; in CPLEX the problem is smaller but, if an user tries to delete one or more intervals of a large number of contiguously indexed variables, then the current code does not take advantage of the fact CPLEX is optimized to do this natively. The proposed change creates the smallest list of intervals and then call delete on them in reverse order (just to minimize the variable block shifts inside CPLEX, if they are done, but also because this simplifies the code that compute the correspondence between indexes).

### DIFF
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -501,7 +501,7 @@ function intervalize(xs)
 end
 
 function MOI.delete(model::Optimizer, indices::Vector{<:MOI.VariableIndex})
-    info = _info.(model, v)
+    info = [_info(model, var_idx) for var_idx in indices]
     soc_idx = findfirst(e -> e.num_soc_constraints > 0, info)
     soc_idx !== nothing && throw(MOI.DeleteNotAllowed(indices[soc_idx]))
     sorted_del_cols = sort!(collect(i.column for i in info))
@@ -509,7 +509,9 @@ function MOI.delete(model::Optimizer, indices::Vector{<:MOI.VariableIndex})
     for ri in reverse(1:length(starts))
         CPLEX.c_api_delcols(model.inner, Cint(starts[ri]), Cint(ends[ri]))
     end
-    delete!.(model.variable_info, indices)
+    for var_idx in indices
+      delete!(model.variable_info, var_idx)
+    end
     for other_info in values(model.variable_info)
         other_info.column -= searchsortedlast(
           sorted_del_cols, other_info.column
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -485,16 +485,19 @@ function MOI.is_valid(model::Optimizer, v::MOI.VariableIndex)
     return haskey(model.variable_info, v)
 end
 
-# expect sequence to be sorted
+# Helper function used inside MOI.delete (vector version). Takes a list of
+# numbers (MOI.VariableIndex) sorted by increasing values, return two lists
+# representing the same set of numbers but in the form of intervals.
+# Ex.: intervalize([1, 3, 4, 5, 8, 10, 11]) -> ([1, 3, 8, 10], [1, 5, 8, 11])
 function intervalize(xs)
     starts, ends = empty(xs), empty(xs)
     for x in xs
-      if isempty(starts) || x != last(ends) + 1
-        push!(starts, x)
-        push!(ends, x)
-      else
-        ends[end] = x
-      end
+        if isempty(starts) || x != last(ends) + 1
+            push!(starts, x)
+            push!(ends, x)
+        else
+            ends[end] = x
+        end
     end
 
     return starts, ends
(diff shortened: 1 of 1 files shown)


odow:
It would be worth benchmarking the difference between this and `MOI.delete.(model, vars)`. Is it much faster? CPLEX doesn't have the lazy update semantic issue that Gurobi has.

henriquebecker91:
Maybe I am doing something wrong, as I did a very crude test, but the results are better than I expected. Seems like using the unspecialized version has some allocation problem.
```julia
import MathOptInterface, CPLEX
const MOI = MathOptInterface

function clean_model_and_vars()
    model = CPLEX.Optimizer()
    vars = MOI.add_variables(model, 10000)
    model, vars
end

# copied from: https://github.com/JuliaOpt/MathOptInterface.jl/blob/c4beceb4e9444c752f7e655f7a50a38a70181b06/src/indextypes.jl#L124-L128
function unspecialized_delete(model::MOI.ModelLike, indices::Vector{<:MOI.Index})
    for index in indices
        MOI.delete(model, index)
    end
end

function benchmark()
    testcases = [
        # all variables
        [1:10000],
        # 20% of the middle variables
        [4000:6000],
        # first 10% last 10%
        [1:1000, 9000:10000],
        # 10% in ten 1% contiguous blocks
        [1:100, 1001:1100, 2001:2100, 3001:3100, 4001:4100,
            5001:5100, 6001:6100, 7001:7100, 8001:8100, 9001:9100],
        # should not be a good case, as it is each other variable
        [1:2:10000],
        # the reverse of the above
        [9999:-2:1]
    ]
    for test in testcases
        indexes2del = collect(Base.Iterators.flatten(collect.(test)))
        @show test
        m, vars = clean_model_and_vars()
        vars2del = vars[indexes2del]
        @time MOI.delete(m, vars2del)
        m_, vars_ = clean_model_and_vars()
        vars2del_ = vars_[indexes2del]
        @time unspecialized_delete(m_, vars2del_)
        @assert MOI.is_valid.(m, vars2del) == MOI.is_valid.(m_, vars2del_)
    end
end

benchmark()
println("second run")
benchmark()
```

And the output after "second run" (in my machine) is:
```
test = StepRange{Int64,Int64}[1:1:10000]
  0.001037 seconds (54 allocations: 756.500 KiB)
  1.378340 seconds (50.06 M allocations: 2.849 GiB, 9.99% gc time)
test = StepRange{Int64,Int64}[4000:1:6000]
  0.000966 seconds (8.06 k allocations: 1.046 MiB)
  0.491413 seconds (18.02 M allocations: 950.539 MiB, 9.35% gc time)
test = StepRange{Int64,Int64}[1:1:1000, 9000:1:10000]
  0.002110 seconds (8.06 k allocations: 1.046 MiB, 57.43% gc time)
  0.501812 seconds (18.02 M allocations: 950.539 MiB, 9.10% gc time)
test = StepRange{Int64,Int64}[1:1:100, 1001:1:1100, 2001:1:2100, 3001:1:3100, 4001:1:4100, 5001:1:5100, 6001:1:6100, 7001:1:7100, 8001:1:8100, 9001:1:9100]
  0.001077 seconds (9.06 k allocations: 1.077 MiB)
  0.268387 seconds (9.51 M allocations: 498.236 MiB, 9.64% gc time)
test = StepRange{Int64,Int64}[1:2:9999]
  0.059585 seconds (5.08 k allocations: 1.205 MiB)
  1.058221 seconds (37.53 M allocations: 1.983 GiB, 9.86% gc time)
test = StepRange{Int64,Int64}[9999:-2:1]
  0.061747 seconds (5.08 k allocations: 1.205 MiB)
  1.014600 seconds (37.53 M allocations: 1.983 GiB, 10.24% gc time)
```
The first time (of my specialized version) is always better than the second time (of the unspecialized version), often by two orders of magnitude.

src/MOI/MOI_wrapper.jl
```diff
@@ -485,6 +485,45 @@ function MOI.is_valid(model::Optimizer, v::MOI.VariableIndex)
     return haskey(model.variable_info, v)
 end
 
+# expect sequence to be sorted
+function intervalize(xs)
+    starts, ends = empty(xs), empty(xs)
+    for x in xs
+      if isempty(starts) || x != last(ends) + 1
```
odow: Please use 4-space indentation

henriquebecker91: Fixed.

henriquebecker91:
I have fixed the indentation and added some comments (for clarifying the utility of the helper function and the source of the performance boost).

Looking back at the code I remembered where the efficiency problem was. While using the native CPLEX function for deleting an interval is also nice, I think the main source of efficiency in the new code is that the shifting of the column indexes in the old code is done in a way that is O(n * m) while in the new code is O(n*log(m)), for 'n' as the number of model variables after deletion and 'm' as the number of variables deleted. If the number of variables deleted is significant (128~1024 variables), even if not contiguous, this is already a considerable speed-up (as it decays from 128~1024 to 7~10, a speed-up of a factor 18~102).

odow:
Thanks @henriquebecker91

PR #275, opened 2020-01-07
Add missing supports for Callbacks

odow:
x-ref https://github.com/JuliaOpt/Gurobi.jl/pull/283


odow:
Not sure why coverage didn't catch the lines inside the callbacks being hit, because they are covered...

PR #277, opened 2020-01-12
Indicator constraints in MOI

matbesancon:
Supports indicator constraints. The removal of these is not yet in the exposed C API (even though it is in CPLEX documentation), so it is simple additions for now. A new dict keeps track of the added indicator constraints


matbesancon:
The tests are currently erroring. Checking for example `MOIT.indicator1_test`, it seems the indicator constraints are not registered. There is unfortunately no way of seeing this with CPLEX.

odow:
> Indicator constraints are not registered. There is unfortunately no way of seeing this with CPLEX.

What do you mean by this?

matbesancon:
The addition of the indicator constraints do not return any error, but the problem result is as is they were not added

odow:
> but the problem result is as is they were not added

You probably need to set `model.has_int = true` in `add_indicator_constraint` since CPLEX has different discrete and continuous solution routines: https://github.com/JuliaOpt/CPLEX.jl/blob/2fe973549f0bba4f7b011fa07598185e34736ea1/src/cpx_solve.jl#L3

matbesancon:
ok the solution was found in tears and blood, there was an off-by-one error in the cpx call, see the fix in `src/cpx_constrs.jl`. Also, the CPLEX documentation uses an Int32 for the sense instead of Char.

All indicator tests are now passing

matbesancon:
Regarding `model.has_int = true`, it is supposed to be set when the binary variable is added. It is up to the user in all solvers to check that the indicator variable provided has been set to bin or integer and bounded

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+    if info_tuple === nothing
+        return false
+    end
+    return typeof(info_tuple[1].set) == S
+end
+
+function MOI.get(model::Optimizer,
+                 ::MOI.ConstraintSet,
+                 c::MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, MOI.IndicatorSet{A, S}}
+) where {A, S}
```
odow: Sorry to be a pain, but can we have consistent formatting? Also, we prefer to drop type parameters if not needed in the function.
```Julia
function MOI.get(
    model::Optimizer,
    ::MOI.ConstraintSet,
    c::MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, <:MOI.IndicatorSet}
)
```

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+    MOI.throw_if_not_valid(model, c)
+    info = model.indicator_constraint_info[c.value][1]
+    return info.set
+end
+
+function MOI.get(model::Optimizer,
+                 ::MOI.ConstraintFunction,
+                 c::MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, MOI.IndicatorSet{A, S}}
+) where {A, S}
```
odow: Ditto formatting

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+    MOI.throw_if_not_valid(model, c)
+    func = model.indicator_constraint_info[c.value][2]
+    return func
+end
+
+function MOI.add_constraint(
+        model::Optimizer, func::MOI.VectorAffineFunction{Float64},
+        s::MOI.IndicatorSet{A, S}) where {A, S <: Union{MOI.GreaterThan{Float64}, MOI.LessThan{Float64}, MOI.EqualTo{Float64}}}
+
+
```
odow:
```Julia
function MOI.add_constraint(
    model::Optimizer, 
    func::MOI.VectorAffineFunction{Float64},
    s::MOI.IndicatorSet{A, S}
) where {
    A, 
    S <: Union{MOI.GreaterThan{Float64}, MOI.LessThan{Float64}, MOI.EqualTo{Float64}}
}
```

And remove the redundant empty lines.

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+    first_index_terms  = [v.scalar_term for v in func.terms if v.output_index == 1]
+    scalar_index_terms = [v.scalar_term for v in func.terms if v.output_index != 1]
+    length(first_index_terms) == 1 || error("There should be exactly one term in output_index 1, found $(length(first_index_terms))")
```
odow: Prefer 
```Julia
if length(first_index_terms) != 1
    error()
end
```

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+            nindices += 1
+        end
+    end
+    return nindices
+end
+
+function MOI.get(model::Optimizer,
+                 ::MOI.ConstraintName,
+                 c::MOI.ConstraintIndex{<:MOI.VectorAffineFunction, S}
+) where {S <: MOI.IndicatorSet}
```
odow: Formatting. Also Prefer `MOI.ConstraintIndex{<:MOI.VectorAffineFunction, <:MOI.IndicatorSet}`.

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+    MOI.throw_if_not_valid(model, c)
+    info = model.indicator_constraint_info[c.value][1]
+    return info.name
+end
+
+function MOI.set(model::Optimizer,
+                 ::MOI.ConstraintName,
+                 c::MOI.ConstraintIndex{<:MOI.VectorAffineFunction, S},
+                 name::String,
+) where {S <: MOI.IndicatorSet}
```
odow: Formatting.

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+    MOI.throw_if_not_valid(model, c)
+    info = model.indicator_constraint_info[c.value][1]
+    info.name = name
+    return nothing
```
odow: JuMP style guide prefers `return` over `return nothing`.

test/MathOptInterface/MOI_wrapper.jl
```diff
@@ -76,10 +80,10 @@ end
 end
 
 @testset "Integer Linear tests" begin
-    MOIT.intlineartest(BRIDGED_OPTIMIZER, CONFIG, [
-        # TODO(odow): Indicator sets not supported.
-        "indicator1", "indicator2", "indicator3", "indicator4"
-    ])
+    # interval somehow needed for indicator tests
+    interval_optimizer = MOIB.LazyBridgeOptimizer(OPTIMIZER)
+    MOIB.add_bridge(interval_optimizer, MOIB.Constraint.SplitIntervalBridge{Float64})
```
odow: Remove the above three lines. We just need `BRIDGED_OPTIMIZER` below.

test/MathOptInterface/MOI_wrapper.jl
```diff
@@ -76,10 +80,10 @@ end
...
+    MOIT.intlineartest(BRIDGED_OPTIMIZER, CONFIG, String[])
```
odow: Remove unneeded `String[]` as final argument.

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+function MOI.get(
+    model::Optimizer, ::MOI.ListOfConstraintIndices{MOI.VectorAffineFunction{Float64}, S}
+) where {S <: MOI.IndicatorSet}
+    indices = MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, S}[]
+    for (key, info_func) in model.indicator_constraint_info
+        info = info_func[1]
+        if typeof(info.set) === S
+            push!(indices, MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, S}(key))
+        end
+    end
```
odow: Alternatively:
```Julia
indices = MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, S}[
    MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, S}(key)
    for (key, info_func) in model.indicator_constraint_info
    if typeof(info_func[1].set) == s
]

File: src/MOI/indicator_constraint.jl [hunk 11]
```diff
@@ -0,0 +1,117 @@
+
+function MOI.supports_constraint(
+    ::Optimizer, ::Type{MOI.VectorAffineFunction{Float64}}, ::Type{<:MOI.IndicatorSet{A, S}}
+) where {A, S <: Union{MOI.LessThan{Float64}, MOI.GreaterThan{Float64}, MOI.EqualTo{Float64}}}
```
odow: Is `EqualTo` tested? 

Is there are a restriction on `A`?

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,117 @@
...
+    return true
+end
+
+function MOI.is_valid(
+    model::Optimizer,
+    c::MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, S}
+) where {A, S <: MOI.IndicatorSet}
```
odow: Is this correct? What is `A` here?

src/MOI/indicator_constraint.jl
matbesancon: what formatting do you refer to, the parenthesis on a new line?

src/MOI/indicator_constraint.jl
matbesancon: We need the exact type of `S` to construct `indices = MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, S}[]` below

odow: Not in this function?

src/MOI/indicator_constraint.jl
odow: Dropping arguments to a new line and indenting 4 spaces, with parenthesis on new line.

test/MathOptInterface/MOI_wrapper.jl
matbesancon: `BRIDGED_OPTIMIZER` is using all bridges, including the bridge from indicator to SOS, the test could thus pass without testing the real implementation if an error is made (like `MOI.supports_constraint` returning the wrong type

odow: Okay, but then we need to test the indicator tests with `interval_optimizer`. Currently, it is un-used.

src/MOI/indicator_constraint.jl
matbesancon: See the docstring for `MOI.IndicatorSet{A, S}`
A is a value from the enum `MOI.ActivationCondition`.:
```
julia> typeof(MOI.ACTIVATE_ON_ONE)
Enum MathOptInterface.ActivationCondition:
ACTIVATE_ON_ZERO = 0
ACTIVATE_ON_ONE = 1
```

Unfortunately, one cannot type-check value parameters of a type

matbesancon: > Is EqualTo tested?

No that's a good point, it is missing from MOI, I'll add a PR with an equality test

src/MOI/indicator_constraint.jl
matbesancon: same as above, `A` is the MOI.ActivationCondition

src/MOI/indicator_constraint.jl
matbesancon: correct, I'm updating that too

src/MOI/indicator_constraint.jl
odow: But `A` isn't used in the function, or in the header. You probably just mean `) where {S <: IndicatorSet}`?

matbesancon: that's right, I hadn't caught this one thanks

test/MathOptInterface/MOI_wrapper.jl
matbesancon: totally correct, just fixed that

src/MOI/indicator_constraint.jl
matbesancon: done

src/MOI/indicator_constraint.jl
matbesancon: addressed in last commit

matbesancon:
bump there

src/MOI/MOI_wrapper.jl
```diff
@@ -305,6 +309,8 @@ function MOI.supports_constraint(
     # Note: CPLEX does not support quadratic equality constraints.
     return true
 end
+import MathOptInterface
+const MOI = MathOptInterface
```
odow: Remove.

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,138 @@
...
+    # a^T x + b <= c <==> a^T x <= c - b
+    fcons = MOI.constant(func)
+    rhs = Cdouble(MOI.constant(s.set) - fcons[2])
+    add_indicator_constraint(model.inner, linear_idx, linear_coefficients, sense_code, rhs, var_inner, active_bool)
+    return MOI.ConstraintIndex{typeof(func), typeof(s)}(model.last_constraint_index)
+end
+
+_get_indicator_sense(::MOI.GreaterThan) = Cint('G')
+_get_indicator_sense(::MOI.LessThan) = Cint('L')
+_get_indicator_sense(::MOI.EqualTo) = Cint('E')
```
odow: Is `Cint` correct here?!? Not `Cchar`?

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,138 @@
...
+
+function MOI.get(
+        model::Optimizer,
+        ::MOI.ListOfConstraintIndices{<:MOI.VectorAffineFunction, S},
+) where {
+    S <: MOI.IndicatorSet
+}
+
```
odow: Remove blank line

src/MOI/indicator_constraint.jl
```diff
@@ -0,0 +1,138 @@
...
+    ::Type{<:MOI.IndicatorSet{A, S}}
+) where {
+    A, S <: Union{MOI.LessThan{Float64}, MOI.GreaterThan{Float64}, MOI.EqualTo{Float64}}
+}
+    return true
+end
+
+function MOI.is_valid(
+        model::Optimizer,
+        c::MOI.ConstraintIndex{MOI.VectorAffineFunction{Float64}, S}
```
odow: Here and throughout, indentation is off.

src/cpx_constrs.jl
```diff
@@ -494,12 +494,12 @@ function add_indicator_constraint(model::Model, idx::Vector{Cint}, coeff::Vector
                       Cint,
                       Cint,
                       Cdouble,
-                      Cchar,
+                      Cint,
```
odow: I understand that the [docs say this is a `Cint`](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.10.0/ilog.odms.cplex.help/refcallablelibrary/gcapi/addindconstr.html), but this is surely a mistake. What would this be the only place in the API where the sense is a `Cint`. What happens if you make this a `Cchar`? Does it work? 

I can follow up with the developers.

src/MOI/indicator_constraint.jl
matbesancon: Yes, I don't know why the signature in the wrapper takes a Cchar:
https://www.ibm.com/support/knowledgecenter/SSSA5P_12.9.0/ilog.odms.cplex.help/refcallablelibrary/gcapi/addindconstr.html
this is specified as int, might have been a char in a past version

src/cpx_constrs.jl
matbesancon: I'll re-test with Cchar to see if it works

odow: I've mailed the CPLEX developers for clarification.

matbesancon: seems all fine, I reverted to Cchar

src/MOI/indicator_constraint.jl
matbesancon: changed in last commit

src/MOI/indicator_constraint.jl
matbesancon: Converted to Cchar in last commit

src/MOI/MOI_wrapper.jl
matbesancon: done in last commit

src/cpx_constrs.jl
odow: Here's the response from CPLEX:

image

Please change to `Cint` and leave a comment.

matbesancon: done thanks for reaching out to them

matbesancon:
this is all good on my side, I think the MOI incompatibility should be fixed in another PR

matbesancon:
the tests are passing with MOI v0.9.11

matbesancon:
bump @odow

odow:
Screenshot of the tests passing?

matbesancon:
image

image

Issue #279, opened 2020-01-13
Tests on callbacks error out


matbesancon:
Runnnig the tests (with #278 fixed)
```julia
HeuristicSolution: Error During Test at /home/mbesancon/Documents/crptests/project_julia/CPLEX.jl/test/MathOptInterface/MOI_callbacks.jl:227
  Test threw exception
  Expression: MOI.submit(model, MOI.HeuristicSolution(cb_data), x, floor.(x_vals)) == MOI.HEURISTIC_SOLUTION_UNKNOWN
  CPLEX.CplexError(1003, "CPLEX Error  1003: Bad argument to Callable Library routine.\n")
  Stacktrace:
   [1] return_status_or_throw at /home/mbesancon/Documents/crptests/project_julia/CPLEX.jl/src/cpx_generic_callbacks.jl:24 [inlined]
   [2] cbpostheursoln at /home/mbesancon/Documents/crptests/project_julia/CPLEX.jl/src/cpx_generic_callbacks.jl:121 [inlined]
   [3] submit(::CPLEX.Optimizer, ::MathOptInterface.HeuristicSolution{CPLEX.CallbackContext}, ::Array{MathOptInterface.VariableIndex,1}, ::Array{Float64,1}) at /home/mbesancon/Documents/crptests/project_julia/CPLEX.jl/src/MOI/MOI_callbacks.jl:219
```

The failing tests are the following:
```julia
Test Summary:                   | Pass  Error  Broken  Total
MathOptInterface               | 2707      4       1   2712
  MOI_callbacks.jl               |   46      4             50
    LazyConstraintCallback |   16                    16
    UserCutCallback            |    6                     6
    HeuristicCallback           |    6      2              8
      HeuristicSolution          |    4      2              6
      LazyConstraint             |    1                     1
      UserCut                        |    1                     1
    CPLEX.CallbackFunction |   18      2             20
      OptimizeInProgress      |   13                    13
      LazyConstraint              |    3                     3
      UserCut                         |    1                     1
      HeuristicSolution           |    1      2              3
```

odow:
Which version of CPLEX do you have

matbesancon:
v12.8.0

odow:
I've only tested 12.9.

matbesancon:
I tested on 12.9 (with the same state of the Julia package), I can confirm the problem is gone. Maybe we should indicate in the README that the callbacks are CPLEX >= v12.9

odow:
[...]

PR #280, opened 2020-01-15
Update README with parameters

odow:


odow:
Uh oh. Looks like we found our bug!

https://travis-ci.com/JuliaOpt/CPLEX.jl/jobs/276846929#L369

matbesancon:
Indeed there it goes, I assume you didn't do anything special?

odow:
No. I just modified the README! Didn't touch any code in that last commit. That definitely suggests either (i) we're using the C API incorrectly; or (ii) there's a bug in the C API.

x-ref https://github.com/JuliaOpt/CPLEX.jl/issues/279#issuecomment-575762967

PR #286, opened 2020-03-02
Update MOI_callbacks.jl

odow:
x-ref: https://discourse.julialang.org/t/jump-0-20-1-possible-bug-in-cplex-when-using-generic-lazy-callback/35425

### DIFF
--- a/src/MOI/MOI_callbacks.jl
+++ b/src/MOI/MOI_callbacks.jl
@@ -88,11 +88,7 @@ function default_moi_callback(model::Optimizer)
             model.lazy_callback(cb_data)
         end
     elseif cb_where == CPX_CALLBACKCONTEXT_RELAXATION
-        callbackgetrelaxationpoint(model, cb_data, cb_where)
-            if model.lazy_callback !== nothing
-                model.callback_state = CB_LAZY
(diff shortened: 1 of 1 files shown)


ryancorywright:
Thanks for looking into this. I applied the patch and reran the code mentioned in the discourse page: CPLEX doesn't throw an error any more, but it rejects every feasible solution and eventually says that the problem is infeasible. 

From the docs https://www.ibm.com/support/knowledgecenter/SSSA5P_12.10.0/ilog.odms.cplex.help/refcallablelibrary/macros/CPX_CALLBACKCONTEXT_RELAXATION.html

"CPLEX invokes the generic callback in this context when it has found a relaxed solution available. The relaxed solution is usually not integer feasible. It can, for example, be the solution to a node LP (but can also come from another place)."

odow:
On this branch we only call the lazy constraint from `CPX_CALLBACKCONTEXT_CANDIDATE`:
https://github.com/JuliaOpt/CPLEX.jl/blob/19c82b917b173189b45a4818331d53dee86a0e8d/src/MOI/MOI_callbacks.jl#L81-L89

Please provide a reproducible example.

PR #291, opened 2020-03-20
Fix handling of discrete->continuous models

### LINKED ISSUE #290: `has_duals` answers differently if variables were originally integers
@odow suggested that I opened a PR here after my [post in Discourse](https://discourse.julialang.org/t/cplex-model-answer-has-duals-differently-if-variables-were-integer/36276). Basically:

```julia
using JuMP
import CPLEX

function test_duals(solver_sym, relaxed_mip)
    solver_lib = getfield(Main, solver_sym)
    m = direct_model(solver_lib.Optimizer())
    set_silent(m)
    if relaxed_mip
        @variable(m, x[1:10] >= 0, Int)
    else
        @variable(m, x[1:10] >= 0)
    end
    @objective(m, Min, sum(x))
    relaxed_mip && unset_integer.(x)
    optimize!(m)
    @show solver_sym
    @show relaxed_mip
    @show has_duals(m)
end

test_duals(:CPLEX, false)
test_duals(:CPLEX, true)
```

Gives `true` for the first case and `false` for the second case, in other words, if the model was a MIP before becoming an LP then the duals do not become available at the end of the LP optimization (both GLPK and Gurobi return `true` in both cases, not just the first). It seems this is not expected behavior but instead it is #110 re-surfacing. I will give it a look, but do not guarantee it will be soon. I need CPLEX working in this case but not now (for now I can just use Gurobi).

odow:
Closes #290

### DIFF
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -1861,6 +1861,15 @@ function _make_problem_type_continuous(optimizer::Optimizer)
     return
 end
 
+function _has_discrete_variables(model::Optimizer)
+    for v in values(model.variable_info)
+        if v.type != CONTINUOUS
+            return true
+        end
+    end
+    return length(model.sos_constraint_info) > 0
+end
+
 function MOI.optimize!(model::Optimizer)
     # Initialize callbacks if necessary.
     if check_moi_callback_validity(model)
@@ -1868,7 +1877,7 @@ function MOI.optimize!(model::Optimizer)
         model.has_generic_callback = false
     end
     model.cached_solution = nothing
-    if model.inner.has_int
+    if _has_discrete_variables(model)
         _make_problem_type_integer(model)
         varindices = Cint[]
         values = Float64[]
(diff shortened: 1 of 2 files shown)


henriquebecker91:
Should [`has_int`](https://github.com/JuliaOpt/CPLEX.jl/blob/009be0f0dc571e8de8f0c56d52b4055c286bf64b/src/cpx_model.jl#L4) keep existing if a complete check will be done each time?

Seems to me that either: (1) There should be a `num_ilp_vars` (counts the number of binary and integer variables) that is always keep updated (need to be changed in `add`, `delete`, `set_{binary,integer}`, `unset_{binary,integer}`). This adds a lot of code and is very bug-prone, as any new methods or change to existing method need to be aware of its existence. However it is `O(1)` to check when necessary. Strangely, looking at the search results for this repository, it seems that setting this variable is far more common than checking it. (2) do what odow has suggested, which gives a hidden cost to `optimize!` (but is not so bad, as `optimize!` will probably touch all variables anyway) but then also seems wise to remove such cached info.

henriquebecker91:
@odow, can I help some way? If this is fixed fast, maybe I will be able to use CPLEX in my next batch of experiments without having to implement a workaround.

odow:
`has_int` needs to stay for MPB etc. This is fine for a near-term fix. When I get some time, maybe this summer, I want to rebuild CPLEX.jl based on Clang.jl.

henriquebecker91:
Great. I just need that the model give the duals when solved relaxed. I just pointed the `has_int` because I did not want you to fix this in a hurry because of me and leave some loose ends.

PR #295, opened 2020-04-06
Prototype of quadratic constraint duality

metab0t:
I add a prototype for computing dual multiplier for quadratic constraint based on `qcpdual.c` official example bundled with CPLEX.

### DIFF
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -2137,15 +2137,56 @@ function MOI.get(
     return _dual_multiplier(model) * model.cached_solution.linear_dual[row]
 end
 
-# function MOI.get(
-#     model::Optimizer, attr::MOI.ConstraintDual,
-#     c::MOI.ConstraintIndex{MOI.ScalarQuadraticFunction{Float64}, <:Any}
-# )
-#     _throw_if_optimize_in_progress(model, attr)
-#     MOI.check_result_index_bounds(model, attr)
-#     pi = model.cached_solution.quadratic_dual[_info(model, c).row]
-#     return _dual_multiplier(model) * pi
-# end
+function MOI.get(
+    model::Optimizer, attr::MOI.ConstraintDual,
+    c::MOI.ConstraintIndex{MOI.ScalarQuadraticFunction{Float64}, <:Any}
+)
+    _throw_if_optimize_in_progress(model, attr)
+    MOI.check_result_index_bounds(model, attr)
+
+    numvar = num_var(model.inner)
+    row = _info(model, c).row
+    slackind, slackval = CPLEX.c_api_getqconstrdslack(model.inner, Cint(row))
+    affine_cols, affine_coefficients, I, J, V = CPLEX.c_api_getqconstr(model.inner, row)
+    x = model.cached_solution.variable_primal
+    denseslack = fill(0.0, numvar)
+    for (ind, val) in zip(slackind, slackval)
+        denseslack[ind + 1] = val
+    end
+
+    #compute The derivative of a quadratic constraint x^TQx + a^Tx + b <= 0 is Q^Tx + Qx + a
+    conetop = true
+    TOL = 1e-6
+    grad = fill(0.0, numvar)
+    for (i, j, v) in zip(I, J, V)
+        grad[i + 1] += v * x[j + 1]
+        grad[j + 1] += v * x[i + 1]
+        if(abs(x[i + 1]) > TOL || abs(x[j + 1]) > TOL)
+            conetop = false
+        end
+    end
+    for (i, v) in zip(affine_cols, affine_coefficients)
+        grad[i + 1] += v
+        if(abs(x[i + 1]) > TOL)
+            conetop = false
+        end
+    end
+
+    #TODO if at top of cone dual multiplier is ill-formed
+    if conetop
+        return 0.0
+    end
+
+    pi = 0.0
+    # choose maximal abs(grad) to compute dual
+    absgradmax_index = argmax(abs.(grad))
+    absgradmax = abs(grad[absgradmax_index])
+    if(absgradmax > TOL)
+        pi = denseslack[absgradmax_index] / grad[absgradmax_index]
+    end
+
+    return _dual_multiplier(model) * pi
+end
 
 function MOI.get(model::Optimizer, attr::MOI.ObjectiveValue)
     _throw_if_optimize_in_progress(model, attr)
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -2770,6 +2770,25 @@ end
 
 function MOI.get(
     model::Optimizer,
+    ::MOI.ConstraintDual,
+    c::MOI.ConstraintIndex{MOI.VectorOfVariables, MOI.SecondOrderCone}
+)
+    row = _info(model, c).row
+    f = MOI.get(model, MOI.ConstraintFunction(), c)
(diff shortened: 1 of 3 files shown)


src/MOI/MOI_wrapper.jl
```diff
@@ -2137,15 +2137,56 @@ function MOI.get(
...
-# )
-#     _throw_if_optimize_in_progress(model, attr)
-#     MOI.check_result_index_bounds(model, attr)
-#     pi = model.cached_solution.quadratic_dual[_info(model, c).row]
-#     return _dual_multiplier(model) * pi
-# end
+function MOI.get(
+    model::Optimizer, attr::MOI.ConstraintDual,
+    c::MOI.ConstraintIndex{MOI.ScalarQuadraticFunction{Float64}, <:Any}
+)
```
odow, line 2143:
```Julia
function MOI.get(
    model::OPtimizer,
    ::MOI.ConstraintPrimal,
    c::MOI.ConstraintIndex{
        MOI.ScalarQuadraticFunction{Float64},
        <: Union{LessThan{Float64}, MOI.GreaterThan{Float64}}
    }
)
```

src/MOI/MOI_wrapper.jl
```diff
@@ -2137,15 +2137,56 @@ function MOI.get(
...
+        end
+    end
+    for (i, v) in zip(affine_cols, affine_coefficients)
+        grad[i + 1] += v
+        if(abs(x[i + 1]) > TOL)
+            conetop = false
+        end
+    end
+
+    #TODO if at top of cone dual multiplier is ill-formed
```
odow, line 2175: What should be done instead?

src/MOI/MOI_wrapper.jl
```diff
@@ -2137,15 +2137,56 @@ function MOI.get(
...
+    if conetop
+        return 0.0
+    end
+
+    pi = 0.0
+    # choose maximal abs(grad) to compute dual
+    absgradmax_index = argmax(abs.(grad))
+    absgradmax = abs(grad[absgradmax_index])
```
odow, line 2183:
```Julia
val, index = findmax(abs.(grad))
```

src/MOI/MOI_wrapper.jl
```diff
@@ -2137,15 +2137,56 @@ function MOI.get(
...
+    end
+
+    #compute The derivative of a quadratic constraint x^TQx + a^Tx + b <= 0 is Q^Tx + Qx + a
+    conetop = true
+    TOL = 1e-6
+    grad = fill(0.0, numvar)
+    for (i, j, v) in zip(I, J, V)
+        grad[i + 1] += v * x[j + 1]
+        grad[j + 1] += v * x[i + 1]
+        if(abs(x[i + 1]) > TOL || abs(x[j + 1]) > TOL)
```
odow, line 2164: Here and below:
```Julia
if abs(x[i + 1]) > TOL || abs(x[j + 1]) > TOL
```

src/MOI/MOI_wrapper.jl
metab0t: Changes done, CPLEX only supports quadratic inequality constraint.

src/MOI/MOI_wrapper.jl
metab0t: In official example, it just reports error and returns. Do we need error handling here?

odow: Do you have a link to the example? It would be good to include here.

test/MathOptInterface/MOI_wrapper.jl
```diff
@@ -95,7 +95,7 @@ end
 
 @testset "Quadratic tests" begin
     # TODO(odow): duals for quadratic problems.
```
odow, line 97: You can remove the TODO :)

src/MOI/MOI_wrapper.jl
odow: Do you have a link to the example? Would be good to include to explain what this function is doing.

metab0t:
cplextest
After update the test file, dual tests seem to be fine and the failed test is probably not related to my patch.

src/MOI/MOI_wrapper.jl
metab0t: I have added a link to CPLEX documentation which contains mathematical explanation and where to find the example code.

test/MathOptInterface/MOI_wrapper.jl
metab0t: Removed

src/MOI/MOI_wrapper.jl
metab0t: I change `0.0` to `NaN` to avoid confusion because NaN may remind users that the dual is ill-formed rather than use 0.0 to continue computation.

metab0t:
soctest
I have added dual for SOC constraint and the failed SOC test is related to `MOI.get(model, MOI.DualStatus())` and `MOI.INFEASIBILITY_CERTIFICATE`. CPLEX cannot provide a dual 
solution.

https://github.com/JuliaOpt/MathOptInterface.jl/blob/1c9c6f791d5b5984ade96b99939d4779da386c76/src/Test/contconic.jl#L1009

blegat:
What is the CPLEX status ? Does CPLEX just not support infeasibility certificates for quadratic problem or is it an issue with this instance ?

metab0t:
The `TerminationStatus` is `MOI.INFEASIBLE` because `CPXgetstat` returns `CPX_STAT_INFEASIBLE`.
Then result `CPXsolninfo` indicates that the current solution is neither primal feasible nor dual feasible so no  infeasibility certificates are given.
I don't know if CPLEX supports infeasibility certificates for quadratic problem generally.

blegat:
Exclude the failing test and run it separately with `duals = false` on the config with a TODO to clarify this

metab0t:
> 
> 
> Exclude the failing test and run it separately with `duals = false` on the config with a TODO to clarify this

Done

test/MathOptInterface/MOI_wrapper.jl
```diff
@@ -115,19 +114,18 @@ end
...
-    ])
-
-    MOIT.soc3test(
-        BRIDGED_OPTIMIZER,
-        MOIT.TestConfig(duals = false, infeas_certificates = false, atol = 1e-3)
-    )
+    MOIT.soctest(BRIDGED_CERTIFICATE_OPTIMIZER, soc_config)
 
-    MOIT.rsoctest(BRIDGED_OPTIMIZER, soc_config)
+    #TODO: rotatedsoc2 needs to disable dual test because CPLEX fails to give INFEASIBILITY_CERTIFICATE
```
blegat, line 121: If I read this in a few months, I won't know if this is an expected behavior of CPLEX or a bug in CPLEX (and then was it reported upstream yet ?).
Write something like
```
rotatedsoc2 needs to disable dual test because CPLEX fails to give INFEASIBILITY_CERTIFICATE
TODO determine whether this is an expected behavior and CPLEX does not support certifying infeasibility of quadratic problems and file it upstream if it is not.
See https://github.com/JuliaOpt/CPLEX.jl/pull/295#issuecomment-610456477
```

metab0t: @blegat Changes applied.

odow:
@metab0t I refactored this PR into #321 and add you as a co-author. Thanks for your help.

metab0t:
@odow Thanks for your refactoring work! The refactored PR LGTM.

PR #299, opened 2020-05-20
Update IIS support to MOI 0.9.14.

dourouc05:
https://github.com/JuliaOpt/MathOptInterface.jl/pull/1056 is merged and will be included in the next MOI release.  

Several return codes are not yet in MOI, though. Should they be included in MOI?

### DIFF
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -2792,24 +2792,24 @@ function MOI.get(model::Optimizer, ::MOI.ConflictStatus)
         return MOI.CONFLICT_FOUND
     elseif model.conflict.stat == CPX_STAT_CONFLICT_FEASIBLE
         return MOI.NO_CONFLICT_EXISTS
-    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_CONTRADICTION # TODO: these codes do not exist in MOI.
-        return MOI.OTHER_LIMIT
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_CONTRADICTION
+        return MOI.NO_CONFLICT_FOUND
     elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_DETTIME_LIM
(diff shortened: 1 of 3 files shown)


src/MOI/MOI_wrapper.jl
```diff
@@ -2799,27 +2783,16 @@ function _ensure_conflict_computed(model::Optimizer)
...
+        return MOI.COMPUTE_CONFLICT_NOT_CALLED
     elseif model.conflict.stat == CPX_STAT_CONFLICT_MINIMAL
-        return MOI.OPTIMAL
+        return MOI.CONFLICT_FOUND
     elseif model.conflict.stat == CPX_STAT_CONFLICT_FEASIBLE
-        return MOI.INFEASIBLE
-    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_CONTRADICTION
+        return MOI.NO_CONFLICT_EXISTS
+    elseif model.conflict.stat == CPX_STAT_CONFLICT_ABORT_CONTRADICTION # TODO: these codes do not exist in MOI.
         return MOI.OTHER_LIMIT
```
odow, line 2796: The return type of this function isn't type-stable. Maybe we need a `MOI.COMPUTE_CONFLICT_FAILED` status. I don't think we need to break out all the possible limits.

dourouc05: I added a few status codes in https://github.com/JuliaOpt/MathOptInterface.jl/pull/1091, without delving into the details.

dourouc05:
Based on https://github.com/JuliaOpt/MathOptInterface.jl/pull/1091, I used `NO_CONFLICT_FOUND` when CPLEX has aborted. 

Is it good enough to be merged?

dourouc05:
With MOI 0.9.14 being tagged, maybe this could get merged?

odow:
Ditto on the screenshot showing tests passing.

dourouc05:
There's a lot of noise in the output, here is the output in two parts: 

image
image

PR #318, opened 2020-10-01
Round bounds for ZeroOne variables to avoid CPLEX warning

### LINKED ISSUE #311: CPLEX---Warning:  Non-integral bounds for integer variables rounded.
Hi,

I got this warning (Non-integral bounds for integer variables rounded) when calling the CPLEX solver.  

If I want to deal with this warning message, how to solve it. 

In addition, I use the CPLEX interface of python to solve the same model, this warning will not appear.

Julia version 1.4.2
CPLEX version 12.10

odow:
Closes #311

### DIFF
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -1432,13 +1432,29 @@ function MOI.add_constraint(
     model::Optimizer, f::MOI.SingleVariable, ::MOI.ZeroOne
 )
     info = _info(model, f.variable)
+    col = Cint(info.column - 1)
+    p_col = Ref(col)
     ret = CPXchgctype(
         model.env,
         model.lp,
         1,
-        Ref{Cint}(info.column - 1),
+        p_col,
         Ref{Cchar}(CPX_BINARY),
     )
+    # Round bounds to avoid the CPLEX warning:
+    #   Warning:  Non-integral bounds for integer variables rounded.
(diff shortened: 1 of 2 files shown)


odow:
Barring objections, I'll leave this open for another 48 hours, then merge.

blegat:
Not sure this is relevant but it seems you are handling cases where both bounds and the ZeroOne constraint are added.
These cases should be an error, see https://github.com/jump-dev/MathOptInterface.jl/blob/0423a6c61c6f47e3baf81642dd60e6c7a3c9e405/src/Utilities/model.jl#L482-L496

odow:
> These cases should be an error

Most definitely not. Where did we discuss this??? People add binary variables with bounds _all_ the time.

blegat:
It seems it was discussed in
https://github.com/jump-dev/MathOptInterface.jl/issues/570#issuecomment-451742785
and then implemented in
https://github.com/jump-dev/MathOptInterface.jl/pull/732/
Looking at the `LOWER_BOUND_MASK` and `UPPER_BOUND_MASK` in more details, it seems that we consider that `ZeroOne` don't set bound (the bounds are 0 and 1, we know it from the existence of the ZeroOne flag in the mask so we don't need to store the bounds anywhere so it does not conflict with other set that need their bounds to be stored.

blegat (approved):

odow:
Correct. `ZeroOne` does not set bounds.

PR #348, opened 2020-12-26
Fix #347

dourouc05:
Current state of the test suite: 

```
Test Summary:                              | Pass  Fail  Error  Broken  Total
MathOptInterface Tests                     | 3456     6     11       1   3474
  test_CPXcallbackabort                    |                 1              1
  test_CallbackFunction_CallbackNodeStatus |                 1              1
  test_CallbackFunction_HeuristicSolution  |                 1              1
  test_CallbackFunction_LazyConstraint     |                 1              1
  test_CallbackFunction_OptimizeInProgress |                 1              1
  test_CallbackFunction_UserCut            |                 1              1
  test_Heuristic                           |    1            1              2
  test_Heuristic_LazyConstraint            |          1                     1
  test_Heuristic_UserCut                   |          1                     1
  test_InterruptException                  |                 1              1
  test_LazyConstraint                      |    1            1              2
  test_LazyConstraint_HeuristicSolution    |          1                     1
  test_LazyConstraint_UserCut              |          1                     1
  test_OptimizeInProgress                  |                 1              1
  test_UserCut                             |    1            1              2
  test_UserCut_HeuristicSolution           |          1                     1
  test_UserCut_LazyConstraint              |          1                     1
  test_CPXPARAM_OptimalityTarget           |    4                           4
  test_ZeroOne_GREATER_THAN                |    4                           4
  test_ZeroOne_INTERVAL                    |    4                           4
  test_ZeroOne_LESS_THAN                   |    4                           4
  test_ZeroOne_NONE                        |    4                           4
  test_automatic_env                       |    1                           1
  test_automatic_env_empty                 |    2                           2
  test_basic_constraint_tests              |  701                         701
  test_conflict_bounds                     |    7                           7
  test_conflict_no_conflict                |    7                           7
  test_conflict_scalaraffine               |    7                           7
  test_conflict_two_bound                  |    9                           9
  test_conflict_two_equalto                |    9                           9
  test_conflict_variables_outside          |   10                          10
  test_cont_int_cont                       |   26                          26
  test_contconic                           |  496                         496
  test_contlineartest                      |  524                         524
  test_contquadratictest                   |  190                         190
  test_copytest                            |   37                          37
  test_default_objective_test              |                            No tests
  test_default_status_test                 |    3                           3
  test_emptytest                           |   12                          12
  test_farkas_dual_max                     |    7                           7
  test_farkas_dual_max_ii                  |    7                           7
  test_farkas_dual_min                     |    7                           7
[...]

PR #349, opened 2020-12-26
Add support for CPLEX 20.1

### LINKED ISSUE #347: Support for CPLEX 20.1?
The new CPLEX has arrived: https://www-01.ibm.com/common/ssi/cgi-bin/ssialias?infotype=AN&subtype=CA&htmlfid=897/ENUS220-474&appname=USN. They changed the versioning style this time... 

I could update the wrapper to support this new version. Modifying `deps/build.jl` doesn't look scary at all, it's just about new file names (like `cplex2010.dll` for Windows), but what about the C functions? I suppose that it's just adding a folder in `src`, like `90` and `gen91` for Gurobi, but the version switch of https://github.com/jump-dev/Gurobi.jl/blob/master/src/Gurobi.jl is not yet implemented, as far as I can tell.

odow:
Closes #348 
Closes #347


odow:
@dourouc05 which version did you download?

PR #350, opened 2020-12-26
Fix local discovery of CPLEX 20.1

metab0t:


metab0t:
I test this on Windows 10 and it works fine but CI fails. 
Multiversion support is hard 😢

dourouc05:
In my previous PR, I had added a third version to look for, on line 35: https://github.com/jump-dev/CPLEX.jl/pull/348/files#diff-708e1c220f34be9ffbf04c0619b1f1a56388096c1df2b95603950d7adc80feaaL35

Also, don't forget to update the messages in the same file.

(This part was missing in #349…)

odow:
Yes, this is my fault. I'll sort it.

dourouc05:
Would it be possible to tag a version with this fix, so that new users don't face the same bug?

PR #358, opened 2021-02-22
Provide a more useful error message to user if no CPLEX binary is found. 

dourouc05:
Instead of displaying a large message that is the same for all platforms, only show the relevant part for the user's platform. Moreover, check if some folder is recognised and propose it if it is. 

Maybe it would be good to indicate if the path that is proposed exists (and the proposed commands should work out of the box) or if the user should check further?

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -39,7 +39,7 @@ function get_error_message_if_not_found()
     end
 
     # Create the error message based on the user's system and the best guess.
-    error_beginning = """
+    return """
     Unable to install CPLEX.jl.
 
     The versions of CPLEX supported by CPLEX.jl are:
@@ -54,23 +54,15 @@ function get_error_message_if_not_found()
     correct location if needed):
     
     ```
-    """
-    
-    error_middle = """
     ENV["CPLEX_STUDIO_BINARIES"] = "$(cplex_studio_folder_guessed)"
     import Pkg
     Pkg.add("CPLEX")
     Pkg.build("CPLEX")
-    """
-    
-    error_end = """
     ```
 
     See the CPLEX.jl README at https://github.com/jump-dev/CPLEX.jl for further
     instructions.
     """
-    
-    return error_beginning * error_middle * error_end
 end
 
 function try_local_installation()
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -20,7 +20,7 @@ function get_error_message_if_not_found()
     # Make the best guess for the CPLEX folder, based on the platform.
     cplex_studio_folders = map(["CPLEX_Studio201", "CPLEX_Studio1210"]) do f
         if Sys.iswindows()
-            return "C:\\\\Program Files\\\\$f\\\\cplex\\\\bin\\\\x86-64_win\\\\"
+            return "C:\\\\Program Files\\\\IBM\\\\ILOG\\\\$f\\\\cplex\\\\bin\\\\x64_win64\\\\"
         elseif Sys.isapple()
             return "/Applications/$f/cplex/bin/x86-64_osx/"
         else
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -18,17 +18,14 @@ end
 
 function get_error_message_if_not_found()
     # Make the best guess for the CPLEX folder, based on the platform.
-    cplex_studio_stem_folders = ["CPLEX_Studio201", "CPLEX_Studio1210"]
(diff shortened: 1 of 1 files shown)


dourouc05:
I must admit I don't understand the error message GitHub gives for the CI. Both commits come from the same fork!?

deps/build.jl
```diff
@@ -16,6 +16,63 @@ function library_name(v)
...
+    """
+    
+    error_end = """
+    ```
+
+    See the CPLEX.jl README at https://github.com/jump-dev/CPLEX.jl for further
+    instructions.
+    """
+    
+    return error_beginning * error_middle * error_end
```
odow, line 73: Why split into three strings only to recombine them?

deps/build.jl
```diff
@@ -16,6 +16,63 @@ function library_name(v)
...
+    cplex_studio_folders_mac = ["/Applications/$f/cplex/bin/x86-64_osx/" for f in cplex_studio_stem_folders]
+    cplex_studio_folders_lin = ["/opt/$f/cplex/bin/x86-64_linux/" for f in cplex_studio_stem_folders]
+    
+    cplex_studio_folders = if Sys.iswindows()
+        cplex_studio_folders_win 
+    elseif Sys.isapple()
+        cplex_studio_folders_mac
+    else
+        cplex_studio_folders_lin
+    end
```
odow, line 32: You could also do
```Julia
cplex_studio_folders = map(["CPLEX_Studio201", "CPLEX_Studio1210"]) do f
    if Sys.iswindows()
        return "C:\\\\Program Files\\\\$f\\\\cplex\\\\bin\\\\x86-64_win\\\\"
    elseif Sys.isapple()
        return "/Applications/$f/cplex/bin/x86-64_osx/"
    else
        return "/opt/$f/cplex/bin/x86-64_linux/"
    end
end
```

odow (commented):
CI only works if you are on JuMP-dev and not a fork.

I can give you `Maintain` rights if you want? Here's what it entails https://jump.dev/pages/governance/#repository-maintainers

Then you'll be able to push to `jump-dev` and trigger CI.

odow:
Any yes, it probably makes sense to try the default argument. No point recommending it if we can make it work.

deps/build.jl
dourouc05: That's a remainder of a previous version of the code, I'm merging the three parts.

deps/build.jl
dourouc05: Thanks, that's much nicer!

deps/build.jl
```diff
@@ -16,6 +16,52 @@ function library_name(v)
     return "$(cpx_prefix)cplex$(v).$(Libdl.dlext)"
 end
 
+function get_error_message_if_not_found()
+    # Make the best guess for the CPLEX folder, based on the platform.
+    cplex_studio_folders = map(["CPLEX_Studio201", "CPLEX_Studio1210"]) do f
+        if Sys.iswindows()
+            return "C:\\\\Program Files\\\\IBM\\\\ILOG\\\\$f\\\\cplex\\\\bin\\\\x64_win64\\\\"
```
odow, line 23: Actually, I think you can go
```Julia
return escape_string("C:\\Program Files\\IBM\\ILOG\\$f\\cplex\\bin\\x64_win64\\")
```

odow:
Did you want Maintain rights?

odow (approved):

dourouc05:
The latest commit also uses the guessed path to check for CPLEX. It's only really tested for Windows, though.

I would be glad to obtain the maintainer rights, it will make things easier to contribute! By the way, maybe something similar can be achieved for CPLEXCP?

odow:
I've sent you an invite.

cc @mlubin @blegat @joaquimg @joehuchette 

Setting up CI requires a secret environment variable that I'm not sure we can distribute outside of JuMP-dev due to license restrictions. What's the status of CPLEXCP? There's a [high bar for transferring repos to JuMP-dev](https://jump.dev/pages/governance/#transferring-repositories-to-jump-dev), but getting automated CI might meet that criteria.

odow:
After you accept the invite, you should close this PR and open a new one on the `jump-dev` remote so we can see the CI run.

dourouc05:
Currently, CPLEXCP is getting outside the state of prototype, with many MOI sets implemented (MILP, MISOCP, and parts of constraint programming). It's far from a 0.1 release, though. 

I'm closing this…

odow:
Let us know when CPLEXCP is ready for a public release, and we can revisit the question of CI.

PR #359, opened 2021-02-22
Provide a more useful error message to user if no CPLEX binary is found. 

dourouc05:
New version of #358, normally CI should work here.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -39,7 +39,7 @@ function get_error_message_if_not_found()
     end
 
     # Create the error message based on the user's system and the best guess.
-    error_beginning = """
+    return """
     Unable to install CPLEX.jl.
 
     The versions of CPLEX supported by CPLEX.jl are:
@@ -54,23 +54,15 @@ function get_error_message_if_not_found()
     correct location if needed):
     
     ```
-    """
-    
-    error_middle = """
     ENV["CPLEX_STUDIO_BINARIES"] = "$(cplex_studio_folder_guessed)"
     import Pkg
     Pkg.add("CPLEX")
     Pkg.build("CPLEX")
-    """
-    
-    error_end = """
     ```
 
     See the CPLEX.jl README at https://github.com/jump-dev/CPLEX.jl for further
     instructions.
     """
-    
-    return error_beginning * error_middle * error_end
 end
 
 function try_local_installation()
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -28,16 +28,6 @@ function possible_paths(cplex_studio_paths::Vector{String})
     end
 end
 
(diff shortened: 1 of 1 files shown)


deps/build.jl
```diff
@@ -16,6 +16,55 @@ function library_name(v)
     return "$(cpx_prefix)cplex$(v).$(Libdl.dlext)"
 end
 
+function possible_paths(cplex_studio_paths::Vector{String})
```
odow, line 19: Since we only ever pass a single string, you probably just want this to be 
`function possible_path(cplex_studio_version::String)`? 

Then you can remove the map in favor of just the `if`.

deps/build.jl
```diff
@@ -52,6 +101,11 @@ function try_local_installation()
                 push!(libnames, joinpath(d, name))
             end
         end
+        
+        guess_worked, guessed_path = guess_cplex_folder(["CPLEX_Studio$v"])
+        if guess_worked
+            push!(libnames, joinpath(guessed_path, name))
+        end
```
odow, line 108:
```Julia
guessed_file = joinpath(possible_path("CPLEX_Studio$v"), name)
if isfile(guessed_file)
    push!(lib names, guessed_file)
end
```

Then you could remove `guess_cplex_folder`.

deps/build.jl
```diff
@@ -16,6 +16,55 @@ function library_name(v)
...
+    * 20.1
+
+    You must download and install one of these versions separately.
+
+    You should set the `CPLEX_STUDIO_BINARIES` environment variable to point to
+    the install location then try again. For example (updating the path to the
+    correct location if needed):
+    
+    ```
+    ENV["CPLEX_STUDIO_BINARIES"] = "$(possible_paths(["CPLEX_Studio201"]))"
```
odow, line 57: `$(possible_path("CPLEX_Studio201")`

odow (commented):
Nice to see CI working!

dourouc05:
> Nice to see CI working!

Indeed, the page is much greener :)!

deps/build.jl
dourouc05: The problem is that `possible_path` takes a vector of possibilities in argument. Here, only the latest one is shown. Making it work with strings would be more complicated, as `map` would iterate over the characters.

odow: I'm suggesting to also modify `possible_path` to take a single string and return a single string.

dourouc05: Indeed, it's no more required to have a vector here. Better this way?

odow (approved):
Please squash and merge when you're happy. (It also gives you the opportunity to edit the commit message.)

image

dourouc05:
This way?

odow:
Perfect!

dourouc05:
Thanks for your support!

Issue #361, opened 2021-03-09
CPLEX 1001: Out of memory and MOI.MEMORY_LIMIT


henriquebecker91:
I recently opened an issue in Gurobi (issue [397](https://github.com/jump-dev/Gurobi.jl/issues/397)) because Gurobi has no configurable memory limits but auto-detect memory exhaustion and throws an exception. The solution (PR [398](https://github.com/jump-dev/Gurobi.jl/pull/398)) was to capture the exception and treat it as the solver has stopped the optimization with a `MOI.MEMORY_LIMIT` status.

[...]

PR #362, opened 2021-03-10
Out-of-memory error now is a termination status.

henriquebecker91:
PR spawned from issue #361. It is basically the same that was done for Gurobi.jl issue [397](https://github.com/jump-dev/Gurobi.jl/issues/397) and PR [398](https://github.com/jump-dev/Gurobi.jl/pull/398). The tests pass:

[](https://postimg.cc/cv2hy9sx)

### DIFF
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -112,20 +112,40 @@ end
 Base.cconvert(::Type{Ptr{Cvoid}}, x::Env) = x
 Base.unsafe_convert(::Type{Ptr{Cvoid}}, env::Env) = env.ptr::Ptr{Cvoid}
 
+function _get_error_string(env::Union{Env, CPXENVptr}, ret::Cint)
+    buffer = Array{Cchar}(undef, CPXMESSAGEBUFSIZE)
+    p = pointer(buffer)
+    return GC.@preserve buffer begin
+        errstr = CPXgeterrorstring(env, ret, p)
+        if errstr == C_NULL
+            "CPLEX Error $(ret): Unknown error code."
+        else
+            unsafe_string(p)
+        end
+    end
+end
+
 function _check_ret(env::Union{Env, CPXENVptr}, ret::Cint)
     if ret == 0
         return
     end
-    buffer = Array{Cchar}(undef, CPXMESSAGEBUFSIZE)
-    p = pointer(buffer)
-    GC.@preserve buffer begin
-        errstr = CPXgeterrorstring(env, ret, p)
-        if errstr == C_NULL
-            error("CPLEX Error $(ret): Unknown error code.")
-        else
-            error(unsafe_string(p))
-        end
+    error(_get_error_string(env, ret))
+end
+
+# If you add a new error code that, when returned by CPLEX inside `optimize!`,
+# should be treated as a TerminationStatus by MOI, to the global `Dict`
+# below, then the rest of the code should pick up on this seamlessly.
+const _ERROR_TO_STATUS = Dict{Cint, MOI.TerminationStatusCode}([
+    # CPLEX Code => TerminationStatus
+    CPXERR_NO_MEMORY => MOI.MEMORY_LIMIT,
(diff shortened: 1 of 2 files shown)


henriquebecker91:
Seems like the CPLEX package in CI is not correctly built?

dourouc05:
You need special access rights to run CI, so it's perfectly normal.

dourouc05:
Would you have a very specific case where this error code happens (maybe tunning parameters)? This way, it could be included in the test suite, but it looks like it will be complicated to reproduce in several versions of CPLEX in the CI environment.

henriquebecker91:
hmmm, not sure if I have something that is easy to reproduce. I have a saved MPS model that will trigger it when we try to solve it relaxed (using `JuMP.read_from_file` + `JuMP.relax_integrality` + `JuMP.optimize!`) with the barrier algorithm. But the `mps.gz` has 200MiB (maybe I could reduce it further by making the variable names single-letter).

dourouc05:
If it's so large, I suppose it also takes quite some time before throwing out an error, I don't know if it's worthwhile to add it to the test suite (it would be much slower to run "just" to catch an edge case).

henriquebecker91:
The problem are twofold:

1. If the environment has a lot of memory, the test will take some time, even if it is some pathological case that makes CPLEX memory grow very fast.
2. If the environment has not a lot of memory, then we risk downsizing the test and it will fail in a test environment with more memory.

I think the only clean solution, if testing is deemed necessary, would involve stubbing/mocking the returned value. So we could instantaneously check if the rest of the code works accordingly.

src/MOI/MOI_wrapper.jl
```diff
@@ -2515,6 +2551,9 @@ end
 
 function MOI.get(model::Optimizer, attr::MOI.RawStatusString)
     _throw_if_optimize_in_progress(model, attr)
+    if haskey(_ERROR_TO_STATUS, model.ret_optimize)
+        return _get_error_string(model.env, model.ret_optimize)
```
odow, line 2555: Have you tested this works when called at anytime? I.e., does it depend on some internal state of the `env`?

odow (approved):
Good by me, modulo the question.

src/MOI/MOI_wrapper.jl
odow:
```Julia
julia> env = CPLEX.Env()
CPLEX.Env(Ptr{Nothing} @0x00007f9c942a1010, false, 0)

julia> CPLEX._check_ret(env, Cint(CPLEX.CPXERR_NO_MEMORY))
ERROR: CPLEX Error  1001: Out of memory.

Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] _check_ret(::CPLEX.Env, ::Int32) at /Users/oscar/.julia/dev/CPLEX/src/MOI/MOI_wrapper.jl:126
 [3] top-level scope at REPL[8]:1
```

What's weird is
```Julia
julia> CPLEX._check_ret(C_NULL, Cint(CPLEX.CPXERR_NO_MEMORY))
ERROR: CPLEX Error  1001: Out of memory.

Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] _check_ret(::Ptr{Nothing}, ::Int32) at /Users/oscar/.julia/dev/CPLEX/src/MOI/MOI_wrapper.jl:126
 [3] top-level scope at REPL[9]:1
```
Why even have the pointer as the argument?

henriquebecker91:
Added the two tests. The tests pass.

[](https://postimg.cc/QBtDYr60)

dourouc05 (approved):

odow (approved):

src/MOI/MOI_wrapper.jl
henriquebecker91: Well, it is developed by IBM. I would guess that not taking `env` would be considered as exposing the internals, because they would be exposing `env` is not needed. This way if they somehow start needing some info on `env` the call already takes the object.

PR #380, opened 2021-11-10
performance issues

Gbrunieau:
Caching CPLEX primal solution to avoid repeated calls to CPXgetx
(largely inspired from what was done on Cbc)

Should fix https://github.com/jump-dev/CPLEX.jl/issues/371

**Before**
```
julia> optimize(Model(CPLEX.Optimizer))
 11.440209 seconds (1.20 M allocations: 22.125 MiB)
2-dimensional DenseAxisArray{Float64,2,...} with index sets:
    Dimension 1, ["a", "b", "c", "d", "e"]
    Dimension 2, Base.OneTo(20000)
```

**After**
```
julia> optimize(Model(CPLEX.Optimizer))
  0.155641 seconds (1.20 M allocations: 22.888 MiB)
2-dimensional DenseAxisArray{Float64,2,...} with index sets:
    Dimension 1, ["a", "b", "c", "d", "e"]
    Dimension 2, Base.OneTo(20000)
```

**Script**
```
using Distributed
using JuMP
using CPLEX
using Profile
function optimize(model)
	NV = 20000
	set = ["a", "b", "c", "d", "e"]
	@variable(model, var[set, 1:NV], lower_bound=0)
	cost = AffExpr(0.0)
	for s in set
		for t in 1:NV
			add_to_expression!(cost, t* var[s, t])
		end
	end
	@objective(model, Min, cost)
	@constraint(model, [s in set, t = 1:NV], var[s, t] >= t)
	set_silent(model)
	optimize!(model)
	@time begin
		value.(var)
	end
end
optimize(Model(CPLEX.Optimizer))
```

### DIFF
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -138,6 +138,9 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     # parameter.
     silent::Bool
 
+    has_solution::Bool
+    variable_primal::Union{Nothing,Vector{Float64}}
+
     # Helpers to remember what objective is currently stored in the model.
     objective_type::_ObjectiveType
     objective_sense::Union{Nothing,MOI.OptimizationSense}
@@ -233,6 +236,9 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
         model.env = env === nothing ? Env() : env
         MOI.set(model, MOI.RawOptimizerAttribute("CPXPARAM_ScreenOutput"), 1)
         model.silent = false
+        model.has_solution = false
+        model.variable_primal = nothing
+    
         model.variable_info =
             CleverDicts.CleverDict{MOI.VariableIndex,_VariableInfo}()
         model.affine_constraint_info = Dict{Int,_ConstraintInfo}()
@@ -303,6 +309,8 @@ function MOI.empty!(model::Optimizer)
     model.user_cut_callback = nothing
     model.heuristic_callback = nothing
     model.generic_callback = nothing
+    model.has_solution = false
+    model.variable_primal = nothing
     return
 end
 
@@ -629,6 +637,10 @@ The C API requires 0-indexed columns.
 """
 function column(model::Optimizer, x::MOI.VariableIndex)
     return _info(model, x).column
+end
+
+function column(model::Optimizer, x::Vector{MOI.VariableIndex})
+    return [_info(model, xi).column for xi in x]
 end
 
 function MOI.add_variable(model::Optimizer)
@@ -2621,6 +2633,7 @@ function MOI.optimize!(model::Optimizer)
     model.solve_time = time() - start_time
     model.has_primal_certificate = false
     model.has_dual_certificate = false
+    model.has_solution = false
     if MOI.get(model, MOI.PrimalStatus()) == MOI.INFEASIBILITY_CERTIFICATE
         resize!(model.certificate, length(model.variable_info))
         ret = CPXgetray(model.env, model.lp, model.certificate)
@@ -2631,6 +2644,10 @@ function MOI.optimize!(model::Optimizer)
         ret = CPXdualfarkas(model.env, model.lp, model.certificate, C_NULL)
         _check_ret(model, ret)
         model.has_dual_certificate = true
+    end
+    model.variable_primal = nothing
+    if MOI.get(model, MOI.ResultCount()) == 1
+        model.has_solution = true
     end
     return
 end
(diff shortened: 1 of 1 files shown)


src/MOI/MOI_wrapper.jl
```diff
@@ -138,6 +138,9 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     # parameter.
     silent::Bool
 
+    has_solution::Bool
```
dourouc05, line 141: Do you really need this member? From what I can see, it is semantically equivalent to `variable_primal === nothing`, is always updated at the same time (except in `MOI.optimize!`), and never read.

src/MOI/MOI_wrapper.jl
```diff
@@ -2801,21 +2818,34 @@ function MOI.get(model::Optimizer, attr::MOI.DualStatus)
...
+
+function _update_cache(model::Optimizer, ::Nothing, f_p::F) where {F}
+    n = length(model.variable_info)
+    x = zeros(n)
+    ret = f_p(model.env, model.lp, x, 0, n-1)
+    _check_ret(model, ret)
+    return x
+end
+
+_get_cached_solution(model::Optimizer, data::Vector{Float64}, x) = data[column(model, x)]
```
dourouc05, line 2831: I don't think that adding this function improves the clarity of the code (given that it's only called once): rather, I would say that it mystifies it a bit when you call this function. I you want to keep it, I would remove the second argument (`data`), as it's always `model.variable_primal`.

src/MOI/MOI_wrapper.jl
```diff
@@ -2801,21 +2818,34 @@ function MOI.get(model::Optimizer, attr::MOI.DualStatus)
...
     end
-    x = Ref{Cdouble}()
-    ret = CPXgetx(model.env, model.lp, x, col, col)
-    _check_ret(model, ret)
-    return x[]
+    model.variable_primal = _update_cache(
+        model,
+        model.variable_primal,
+        CPXgetx,
+        )
```
dourouc05, line 2847: Did you run JuliaFormatter locally? Not really important, but it helps maintain the consistency of the code.

src/MOI/MOI_wrapper.jl
```diff
@@ -2801,21 +2818,34 @@ function MOI.get(model::Optimizer, attr::MOI.DualStatus)
     end
 end
 
+_update_cache(::Optimizer, data::Vector{Float64}, ::Any) = data
+
+function _update_cache(model::Optimizer, ::Nothing, f_p::F) where {F}
+    n = length(model.variable_info)
+    x = zeros(n)
+    ret = f_p(model.env, model.lp, x, 0, n-1)
```
dourouc05, line 2826: Is the parameter `f_p` really required? You're only using `CPXgetx` and it makes the call look strange.

odow:
Did you profile to confirm where the slow-down is coming from?
```Julia
using ProfileView
@profview optimize(Model(CPLEX.Optimizer))
```
It'd be nice to know if the problem is the getx, or something else.

Gbrunieau:
@odow Yes it looks like it's the CPXgetx
Here is the result with 
`@profview value.(var)`

profview

src/MOI/MOI_wrapper.jl
odow: @Gbrunieau you can see the formatting failures here: https://github.com/jump-dev/CPLEX.jl/runs/4169362806?check_suite_focus=true

src/MOI/MOI_wrapper.jl
odow: @Gbrunieau please remove `has_solution`. I just needed that for Cbc because the `ResultCount` check was slow.

src/MOI/MOI_wrapper.jl
odow: Agreed. Please hard-code `CPXgetx`. We can refactor in future if needed.

src/MOI/MOI_wrapper.jl
odow: Agreed Just call `model.variable_primal[column(model, x)]` directly in `MOI.get` below.

src/MOI/MOI_wrapper.jl
Gbrunieau: you're right, thanks, it's not needed, I updated the code

src/MOI/MOI_wrapper.jl
Gbrunieau: done

src/MOI/MOI_wrapper.jl
Gbrunieau: thanks I didn't know about JuliaFormatter, I will use it from now on.

src/MOI/MOI_wrapper.jl
Gbrunieau: yes I was thinking maybe we would need something similar for the dual as well, but for now it's ok to hard-code, it's done.

odow (approved):
Did you run the tests? Don't worry if `test_attribute_SolverVersion` fails: https://github.com/jump-dev/CPLEX.jl/pull/381

Gbrunieau:
> Did you run the tests? Don't worry if `test_attribute_SolverVersion` fails: #381

Yes, and indeed I have one failing (which is the one you mention)

```
Test Summary:                                                                  | Pass  Error  Total
MathOptInterface Tests                                                         | 3316      1   3317
```

odow:
Great. I'll merge #381 to fix the test failure first, and then I'll merge this.

odow:
Thanks!

PR #393, opened 2022-04-01
[Performance] don't pass names by default

odow:
Closes https://github.com/jump-dev/CPLEX.jl/issues/392

### DIFF
--- a/src/MOI/MOI_wrapper.jl
+++ b/src/MOI/MOI_wrapper.jl
@@ -241,6 +241,13 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     heuristic_callback::Union{Nothing,Function}
     generic_callback::Any
 
+    # For more information on why `pass_names` is necessary, read:
+    # https://github.com/jump-dev/CPLEX.jl/issues/392
+    # The underlying problem is that we observed that add_variable, then set
+    # VariableName then add_variable (i.e., what CPLEX in direct-mode does) is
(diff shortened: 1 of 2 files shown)


mlubin (approved):

src/MOI/MOI_wrapper.jl
```diff
@@ -130,7 +130,7 @@ function _check_ret_optimize(model)
 end
 
 """
-    Optimizer(env::Union{Nothing, Env} = nothing)
+    Optimizer(env::Union{Nothing, Env} = nothing; pass_names::Bool = false)
```
joaquimg, line 133: Why not an AbstractOptimizerAttribute?

odow: Because this is really a CPLEX.jl hack, rather than an option supported by CPLEX.

joaquimg: Still, it would make sense to have attributes of the wrapper. I don't consider a hack, but an actual option of the wrapper.

The attribute (with a supports function) path is more generic, for users calling multiple solvers in a more abstract manner.

In fact, Xpress will require this eventually.

blegat: I agree with @joaquimg. It could be `optimizer_with_attributes(CPLEX.Optimizer, CPLEX.PassNames() => true)`

odow: Okay, I'll make a new PR.

PR #394, opened 2022-04-03
Adds annotation example to README.md

kaarthiksundar:


odow:
I cleaned up a few things. Did you test this? Because `cnt` was not defined.

kaarthiksundar:
Yes I tested it. I forgot to remove the count variable.

PR #401, opened 2022-04-26
Add license headers

odow:
The need for this came up in some work that will soon be publicly released. I think we should do it in all the JuMP-dev repos.

cc @lschork2

cc @joehuchette


src/MOI/indicator_constraint.jl
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2013: Joey Huchette and contributors
+#
+# Use of this source code is governed by an MIT-style license that can be found
+# in the LICENSE.md file or at https://opensource.org/licenses/MIT.
+
```
odow, line 6:
```suggestion
```

PR #404, opened 2022-07-24
Add example of using CPLEX C attrs (silence output)

PiotrZakrzewski:
It took me a while to find out how to silence output from CPLEX (useful if you use it as part of a script/application and not interactively) so i thought maybe it could be handy to add it in the README.md, plus it is a nice example to help people understand how to set attributes in general.

I hope it helps.

cheers
Piotr Zakrzewski

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,10 @@ CPLEX.jl has two components:
 The C API can be accessed via `CPLEX.CPXxx` functions, where the names and
 arguments are identical to the C API. See the [CPLEX documentation](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.10.0/COS_KC_home.html)
 for details.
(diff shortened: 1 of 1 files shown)


odow:
Options are covered here: https://github.com/jump-dev/CPLEX.jl#use-with-jump. Is that not sufficient?

odow:
If you're using JuMP, you can also call `set_silent(model)`: https://jump.dev/JuMP.jl/stable/manual/models/#Turn-off-output

PiotrZakrzewski:
> If you're using JuMP, you can also call `set_silent(model)`: https://jump.dev/JuMP.jl/stable/manual/models/#Turn-off-output

this is briliant! Pity I didn't finds it earlier. Thanks, I will close this PR.

PR #405, opened 2022-08-07
Add support for Absolute and Relative gap tolerances

mtanneau:
As far as I can tell, this needs MOI v1.7


odow:
Yeah, we still need to tag v1.7

PR #417, opened 2023-01-23
Support CPLEX v22.1.1 in deps/build.jl

metab0t:


odow (approved):
Ooops. Yeah I forgot to change this part.

deps/build.jl
```diff
@@ -14,6 +14,7 @@ const _CPX_VERS = [ # From oldest to most recent.
     "1210", "12100",
     "201", "2010", "20100",
     "221", "2210", "22100",
+    "2211",
```
odow, line 17: Do we also need `22110`? I don't have it installed so I can't see the naming.

odow, line 17:
```suggestion
    "2211", "22100",
```

PR #433, opened 2023-04-19
Fix ObjectiveFunctionType with unset objective

odow:


odow:
Fix needed was https://github.com/jump-dev/MathOptInterface.jl/pull/2150

